PR #7, opened 2017-01-18
Syncing with SVN version

david-binda:
CheckReturnValueSniff: allowing the function to be checked for multiple callees and adding check for checking get_the_terms and get_the_tags passed to foreach w/o checking the return value.


david-binda:
Merging since this is already being used in production w/o issues.

PR #11, opened 2017-01-19
Report on get_childrent function call due to default -1 for posts_per_page

### LINKED ISSUE #10: Report on get_children w/o setting limit to something else than -1

david-binda:
Since the `get_children` is `setting posts_per_page` to `-1` by default, we should make sure that developers are overriding the default value to something reasonable - eg.: 100

This commit is changing the parent of the `WordPressVIPMinimum_Sniffs_VIP_RestrictedFunctionsSniff` class from `WordPress_AbstractFunctionRestrictionsSniff` to `WordPress_Sniffs_VIP_RestrictedFunctionsSniff` and takes advantage of the inheritance which should allows us to override the original mesages in future.

Further, it is adding the `get_children` group with a message about the default `-1` value for `posts_per_page`.

The changeset also excluded the original WordPress.VIP.RestrictedFunctions since we now have our own with all groups.

Fixes #10

### DIFF
--- a/WordPressVIPMinimum/Sniffs/VIP/RestrictedFunctionsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/VIP/RestrictedFunctionsSniff.php
@@ -32,7 +32,7 @@ class WordPressVIPMinimum_Sniffs_VIP_RestrictedFunctionsSniff extends WordPress_
 			),
 			'get_children' => array(
 				'type' => 'error',
-				'message' => '%s() performs a no-LIMIT query by default, make sure to set a reasonable posts_per_page. %s() will do a -1 query by default, a maximum of 100 should be used.'
+				'message' => '%s() performs a no-LIMIT query by default, make sure to set a reasonable posts_per_page. %s() will do a -1 query by default, a maximum of 100 should be used.',
 				'functions' => array(
 					'get_children',
 				),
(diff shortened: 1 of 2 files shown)


WordPressVIPMinimum/Sniffs/VIP/RestrictedFunctionsSniff.php
```diff
@@ -27,7 +30,17 @@ public function getGroups() {
 				'message' => '%s is prohibited on the WordPress.com VIP platform',
 				'functions' => array( 'get_super_admins' ),
 			),
+			'get_children' => array(
+				'type' => 'error',
+				'message' => '%s() performs a no-LIMIT query by default. Make sure to set a reasonable posts_per_page, get_children will do a -1 query by default, a maximum of 100 should be used.',
```
emrikol: I think the punctuation on this message might be a bit ambiguous.

I suggest something like:

`%s() performs a no-LIMIT query by default, make sure to set a reasonable posts_per_page.  get_children will do a -1 query by default, a maximum of 100 should be used.`

david-binda:
Thanks for the suggestion @emrikol . I have addressed the the punctation feedback in 1718c34a2912a2ea175d422f64c33973a5125f86 and 9e1fc29f1b1860fe7e3f2d28508d2ef531179eef

PR #56, opened 2017-04-05
Adding sniff reporting on VIP in a comment

david-binda:
It should help us to catch situations in which a VIP applied hotfix is being accidentally removed.


david-binda:
This is, of course, not working the way we need. Since we need the PHPCS in the DQ report this on comment removal. I'll have to figure that out.

PR #12, opened 2017-01-20
Merging erorr message for non-validated and not-sanitized input variable if both are true

david-binda:
This commit delays the error adding and merges two errors produced by a single sniff into one, if both are about to be triggered.

Fixes #8

PR #61, opened 2017-04-17
Remove the custom ValidatedSanitizedInput.php Sniff

### LINKED ISSUE #58: VIPCS not working with latest version of WordPressCS
Using the latest `master` version of WordPress Coding Standards with VIP Coding Standards is not working properly with PHP 5.6, throwing a fatal error:

```
PHP Fatal error:  Class WordPressVIPMinimum_Sniffs_VIP_ValidatedSanitizedInputSniff contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (WordPress_Sniff::process_token) in /Users/xxxxxx/wpcs/WordPressVIPMinimum/Sniffs/VIP/ValidatedSanitizedInputSniff.php on line 166
```

My workaround was to revert to a previous commit on WPCS using `git reset --hard b39490465f6fd7375743a395019cd597e12119c9`

david-binda:
And exclude the WordPress.VIP.ValidatedSanitizedInput.MissingUnslash via ruleset.xml

Fixes #58

This is going to introduce a regression for changes introduced in https://github.com/Automattic/VIP-Coding-Standards/pull/12 , but we really should push those upstream and we should be able to live with those for some time. The #58 is more severe, as it's breaking the standard for users.


david-binda:
https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/934

GaryJones:
@david-binda Why was MissingUnslash excluded here?

PR #126, opened 2017-12-15
Adding merge conflict sniff and unit tests

david-binda:


jrfnl:
FYI: There is already a PR (by me) open for the same in PHPCS upstream: https://github.com/squizlabs/PHP_CodeSniffer/pull/1724

jrfnl:
The above mentioned upstream PR has been merged and will be released as part of PHPCS 3.4.0.

I'd like to suggest you run the upstream sniff against your unit tests and your sniff against the unit tests of the upstream sniff to see the differences and possibly consider replacing this sniff with the upstream one instead.

PR #148, opened 2018-02-01
Fix a misleading comment

justnorris:
A very minimalistic PR :)


david-binda:
Thanks for fixing the comment @justnorris !

PR #149, opened 2018-02-02
Show a more helpful message for strip_tags() error

### LINKED ISSUE #133: Suggest using `wp_kses` over `strip_tags` in case `allowable_tags` are being used
Eg.:

```
strip_tags( $some_variable, '<p><a><iframe><strong><img>' );
```

is wrong, as it does not strip all the evil scripts (attributes). `wp_kses` should be used.

This is meant for surpassing `wp_strip_all_tags` suggestion over `strip_tags` when whitelisted tags are needed.

justnorris:
Fixes #133 

I silenced the WordPress sniff for `strip_tags()` and replaced it with a custom sniff. 

This is going to check if `strip_tags()` function call contains a comma (and therefor, multiple arguments) - and if it does - it will change the message accordingly to use `wp_kses()` instead.

The two error messages I came up with are:
```
`strip_tags()` does not strip CSS and JS in between the script and style tags. Use `wp_strip_all_tags()` to strip all tags.
```

```
`strip_tags()` does not strip CSS and JS in between the script and style tags. Use `wp_kses()` instead to allow only the HTML you need.
```

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
@@ -15,7 +15,7 @@ class StripTagsNotAllowedSniff implements \PHP_CodeSniffer_Sniff {
 	 *
 	 * @var array
 	 */
-	private $_tokens = [];
+	private $_tokens = array();
 
 	/**
 	 * The phpcs file.
@@ -50,7 +50,6 @@ class StripTagsNotAllowedSniff implements \PHP_CodeSniffer_Sniff {
 
 		$this->_tokens    = $phpcsFile->getTokens();
 		$this->_phpcsFile = $phpcsFile;
-
 
 		if ( 'strip_tags' !== $this->_tokens[ $stackPtr ]['content'] ) {
 			// We only care for strip_slashes function.
--- a/WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
@@ -1,4 +1,9 @@
 <?php
+/**
+ * WordPressVIPMinimum Coding Standard.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
 
 namespace WordPressVIPMinimum\Sniffs\Functions;
 
--- a/WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
@@ -1,0 +1,111 @@
+<?php
+
+namespace WordPressVIPMinimum\Sniffs\Functions;
+
+use PHP_CodeSniffer_File as File;
+use PHP_CodeSniffer_Tokens as Tokens;
+
+/**
+ * This sniff will check for `strip_slashes()` usage and return a relavant error
+ */
+class StripTagsNotAllowedSniff implements \PHP_CodeSniffer_Sniff {
+
+	/**
+	 * Tokens of the whole file.
+	 *
+	 * @var array
+	 */
+	private $_tokens = [];
+
+	/**
+	 * The phpcs file.
+	 *
+	 * @var File
+	 */
+	private $_phpcsFile;
+
+
+	/**
+	 * Returns the token types that this sniff is interested in.
+	 *
+	 * @return array(int)
+	 */
+	public function register() {
+
+		return Tokens::$functionNameTokens;
+
+	}
+
+
+	/**
+	 * Processes the tokens that this sniff is interested in.
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file where the token was found.
+	 * @param int                         $stackPtr  The position in the stack where
+	 *                                               the token was found.
+	 *
+	 * @return bool
+	 */
+	public function process( File $phpcsFile, $stackPtr ) {
+
+		$this->_tokens    = $phpcsFile->getTokens();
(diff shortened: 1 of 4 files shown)


WordPressVIPMinimum/Tests/Functions/StripTagsNotAllowedUnitTest.inc
```diff
@@ -0,0 +1,3 @@
+<?php
+strip_tags( $foo, $bar ); // not-ok
+strip_tags( $foo ); // not-ok
```
GaryJones: How about:

```php
strip_tags( 'Foo, Bar' );
strip_tags( 'Foo' . ', ' . 'Bar' );
strip_tags( 'Foo, Bar', $bar );
```
?

tomjn: Agreed, @justnorris can you make the change?

WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
```diff
@@ -0,0 +1,115 @@
...
+		if ( ! $this->isFunctionCall( $stackPtr ) ) {
+			// Not a function call.
+			return false;
+		}
+
+		if ( $phpcsFile->findNext( T_COMMA, $stackPtr + 1, null, false, null, true ) ) {
+			// If there is a comma inside - there are more than 1 argument to strip_tags().
+			$phpcsFile->addError( '`strip_tags()` does not strip CSS and JS in between the script and style tags. Use `wp_kses()` instead to allow only the HTML you need.', $stackPtr, 'stripTagsNotAllowed' );
+		} else {
+			// `strip_tags()` called with 1 or less arguments.
```
GaryJones: s/less/fewer

WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
```diff
@@ -0,0 +1,115 @@
...
+
+
+	/**
+	 * Check whether the currently examined code is a function call.
+	 *
+	 * @param int $stackPtr The position of the current token in the stack passed in $tokens.
+	 *
+	 * @return bool
+	 */
+	private function isFunctionCall( $stackPtr ) {
```
GaryJones: This seems like something that could be re-used - if PHPCS hasn't got anything already - maybe we could consider a base `Sniff` class like WPCS has.

WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
```diff
@@ -0,0 +1,115 @@
...
+	 *                                               the token was found.
+	 *
+	 * @return bool
+	 */
+	public function process( File $phpcsFile, $stackPtr ) {
+
+		$this->_tokens    = $phpcsFile->getTokens();
+		$this->_phpcsFile = $phpcsFile;
+
+		if ( 'strip_tags' !== $this->_tokens[ $stackPtr ]['content'] ) {
```
GaryJones: Consider using [`get_declaration_name()`](https://github.com/squizlabs/PHP_CodeSniffer/blob/1e440f47e304e640217645e843cdacaacc5339b6/src/Files/File.php#L1219).

WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
```diff
@@ -0,0 +1,115 @@
...
+			// We only care for strip_slashes function.
+			return false;
+		}
+
+		if ( ! $this->isFunctionCall( $stackPtr ) ) {
+			// Not a function call.
+			return false;
+		}
+
+		if ( $phpcsFile->findNext( T_COMMA, $stackPtr + 1, null, false, null, true ) ) {
```
GaryJones: Instead of looking for commas, use [`$this->getMethodParameters(...)`](https://github.com/squizlabs/PHP_CodeSniffer/blob/1e440f47e304e640217645e843cdacaacc5339b6/src/Files/File.php#L1285), and then do a count.

GaryJones:
Relevant: As of [WPCS 1.0.0](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/69ee4c9d75518f8ab8b22c1b9e19885fcd6e8f9e/CHANGELOG.md#changed-1):

> The WordPress.WP.AlternativeFunctions sniff will now only throw a warning if/when the recommended alternative function is available in the minimum supported WP version of a project. In addition to this, certain alternatives are only valid alternatives in certain circumstances, like when the WP version only supports the first parameter of the PHP function it is trying to replace. This will now be taken into account for:
> - wp_strip_all_tags() is only a valid alternative for the PHP native strip_tags() when the second parameter $allowed_tags has not been passed.

WordPressVIPMinimum/Tests/Functions/StripTagsNotAllowedUnitTest.inc
justnorris: I missed this. I'll add this to my todo list.

WordPressVIPMinimum/Sniffs/Functions/StripTagsNotAllowedSniff.php
GaryJones: OK, this is the wrong advice.

`getMethodParameters()` is for getting info about the parameter - whether it's nullable, if it has a type declaration, what the variable name is, whether it is passed by reference, etc.

Much better would be to use the function parameters logic from WPCS. See https://github.com/Automattic/VIP-Coding-Standards/pull/231#issuecomment-430545956 for the links and consequent usage.

rebeccahum:
Closing in favour of #257.

PR #152, opened 2018-02-05
Add new sniff to find dynamic calls to functions

gudmdharalds:
Dynamic calls to certain functions are not allowed in PHP 7.1 and forwards. This sniff will find some of these.

http://php.net/manual/en/migration71.incompatible.php

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
@@ -60,8 +60,7 @@ class DynamicCallsSniff implements \PHP_CodeSniffer_Sniff {
 	 * @return array(int)
 	 */
 	public function register() {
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
```diff
@@ -0,0 +1,271 @@
...
+	/**
+	 * Returns the token types that this sniff is interested in.
+	 *
+	 * We want everything variable- and function-related.
+	 *
+	 * @return array(int)
+	 */
+	public function register() {
+		return array( T_VARIABLE => T_VARIABLE ) +
+			Tokens::$functionNameTokens;
```
david-binda: It does not look like the script is starting of function name tokens. Only variables are being tested. Let's stick with T_VARIABLE in here only.

PR #158, opened 2018-02-21
Sniff to detect create_function()

gudmdharalds:
create_function() is deprecated as of PHP 7.2.0 (see: http://php.net/manual/en/function.create-function.php). This sniff will generate an error when call to that function is detected.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Functions/CreateFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/CreateFunctionSniff.php
@@ -76,7 +76,7 @@ class CreateFunctionSniff implements \PHP_CodeSniffer_Sniff {
 			if ( 'create_function' !==
 				$this->_tokens[ $this->_stackPtr ]['content']
 			) {
-				// Not a function call, repeat.
+				// Not the function call we want.
 				$functionName = $this->getNextTString(
 					$phpcsFile,
 					$functionName + 1
@@ -97,7 +97,7 @@ class CreateFunctionSniff implements \PHP_CodeSniffer_Sniff {
 				( false === $bracket ) ||
 				( T_OPEN_PARENTHESIS !== $this->_tokens[ $bracket ]['code'] )
 			) {
-				// Not a function call, repeat.
+				// Not a function call.
 				$functionName = $this->getNextTString(
 					$phpcsFile,
 					$functionName + 1
--- a/WordPressVIPMinimum/Sniffs/Functions/CreateFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/CreateFunctionSniff.php
@@ -37,6 +37,26 @@ class CreateFunctionSniff implements \PHP_CodeSniffer_Sniff {
 	}
 
 	/**
+	 * Get next T_STRING token.
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file where the token was found.
+	 * @param int                         $stackPtr  The position in the stack where
+	 *                                               the token was found.
+	 *
+	 * @return void
+	 */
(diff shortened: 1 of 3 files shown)


gudmdharalds:
This should solve #2

WordPressVIPMinimum/Tests/Functions/CreateFunctionSniff.inc
```diff
@@ -0,0 +1,16 @@
+<?php
+
+$a = time();
+$b = time() * 2;
+
+$wp_random_testing = create_function( '$a, $b', 'return ( $b / $a ); ');
+
+echo esc_html( $wp_random_testing( $a, $b ) );
+
+
```
GaryJones: All these random extra lines are redundant :-)

WordPressVIPMinimum/Sniffs/Functions/CreateFunctionSniff.php
```diff
@@ -0,0 +1,83 @@
...
+				$this->_tokens[ $this->_stackPtr ]['type'] )
+			||
+			( 'create_function' !==
+				$this->_tokens[ $this->_stackPtr ]['content'] )
+		) {
+			return;
+		}
+
+		$t_item_key = $this->_phpcsFile->findNext(
+			array( T_OPEN_PARENTHESIS ),
```
david-binda: I personally would look up the very next non-whitespace token and checked whether it's a bracket or not. By looking specifically for open bracket within current statement, the `\PHP_CodeSniffer\Files\File::findNext` method may match following code, for instance:

```
class Foo {
    public create_function = 'foo';
}
$c = new Foo();
$c->create_function = 'bar' . ( 1 === 1 ?? 'foo' );
```

Definitely an edge case, but I personally find the approach used eg. in https://github.com/squizlabs/PHP_CodeSniffer/blob/a9d8185ed9a23a981ca8bd698594b67f21a60e90/src/Standards/MySource/Sniffs/PHP/ReturnFunctionValueSniff.php#L47,L50 more robust.

btw: it would be good to add the code I shared above, or similar, to the testing suite, since it might help with further maintenance of this code, as it hardens the implementation of this sniff.

WordPressVIPMinimum/Sniffs/Functions/CreateFunctionSniff.php
```diff
@@ -0,0 +1,119 @@
...
+	 * Processes the tokens that this sniff is interested in.
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file where the token was found.
+	 * @param int                         $stackPtr  The position in the stack where
+	 *                                               the token was found.
+	 *
+	 * @return void
+	 */
+	public function process( File $phpcsFile, $stackPtr ) {
+		$this->_tokens    = $phpcsFile->getTokens();
```
david-binda: I'm not really sure we need to expose `_tokens`, `_phpcsFile` and `_stackPrt` as class properties, as they really are only used in terms of the `process` method (eg.: the `getNextTString` method expects PHP_CodeSniffer_File as it's param, despite it's present in the class property).

WordPressVIPMinimum/Sniffs/Functions/CreateFunctionSniff.php
```diff
@@ -32,88 +32,59 @@ class CreateFunctionSniff implements \PHP_CodeSniffer_Sniff {
...
-			break;
+		if (
+			( false === $bracket ) ||
+			( T_OPEN_PARENTHESIS !== $tokens[ $bracket ]['code'] )
+		) {
+			return;
 		}
+
+		$phpcsFile->addError(
+			'create_function() is deprecated as of ' .
```
GaryJones: This concatenation / line-break seems unnecessary.

PR #161, opened 2018-02-27
Wpcs vip migration

tomjn:
We rely on the VIP ruleset in the WP coding standards, but that's not a repo we control. This PR copies that ruleset into our own

A similar PR is being prepared to remove the original, some coordination will be needed first before it's merged


tomjn:
We're going to need to rename the WordPress-VIP ruleset to prevent clashes with WPCS during the deprecation process

tomjn:
Things have changed since this PR was made, I'd suggest taking a different approach by moving sniff by sniff rather than a whole ruleset

PR #168, opened 2018-03-07
Add the new VIP Go ruleset

jacklenox:


WordPress-VIP-Go/ruleset.xml
```diff
@@ -0,0 +1,86 @@
...
+	</rule>
+
+	<rule ref="WordPressVIPMinimum.VIP.RestrictedFunctions.get_adjacent_post_get_adjacent_post">
+		<type>warning</type>
+	</rule>
+
+	<!-- VIP Uncached warnings -->
+	<rule ref="WordPressVIPMinimum.VIP.RestrictedFunctions.get_page_by_path_get_page_by_path">
+		<type>warning</type>
+		<message>%s() is uncached, please use wpcom_vip_get_page_by_path() instead.</message>
```
david-binda: Is the "uncached" wording specific to VIP Go only? If not, we perhaps should move it to the WordPressVIPMinimum standard's ruleset.

sboisvert:
>just, please, open an issue for the integration tests in case we are not going to act on that bit of feedback now.

@jacklenox can you do this?

WordPress-VIP-Go/ruleset.xml
sboisvert: agreed cc @jacklenox since you're in charge of changing the messages

PR #185, opened 2018-07-16
Add coding standards checks to VIP CS

GaryJones:
Applies most recent WPCS (0.14.1) standards, and PHPCompatibility (testing for PHP 5.3+) to the code in this package. See individual commits for a clearer breakdown.

Couple of comment typos fixed as well :-)

One thing addressed, is the presence of `// end ...` comments. The upstream `CommentedOutCode` sniff was updated a while ago and become more accurate and now throws a number of warnings for `// End ...` comments.

As the rule regarding end comments has been removed from the WP handbook anyway, we may as well remove them all and get rid of the warnings.

Related:
A large number of functions had a blank line at the end of the function for readability what with the end comment after the function closing brace. Those are now superfluous and have been removed as well.


composer.json
```diff
@@ -14,17 +14,22 @@
 		"squizlabs/php_codesniffer": "^2.9.0 || ^3.0.2",
 		"wp-coding-standards/wpcs": "*"
 	},
+	"require-dev" : {
+		"phpcompatibility/php-compatibility": "^8"
```
GaryJones: PHPCompatibility will have breaking changes for `9.0.0`.

composer.json
```diff
@@ -14,17 +14,22 @@
 		"squizlabs/php_codesniffer": "^2.9.0 || ^3.0.2",
 		"wp-coding-standards/wpcs": "*"
```
GaryJones: Also checked with `dev-develop`, which is as close to `1.0.0` as it can be, and no new violations are reported :+1:

.travis.yml
```diff
@@ -48,7 +48,7 @@ matrix:
     - php: 5.3
       env: PHPCS_BRANCH=2.9.0 SHORT_OPEN_TAGS=true
       dist: precise
-    # Run custom integration test. I just picked to it agains PHP 7.0.
+    # Run custom integration test. I just picked to it against PHP 7.0.
```
jrfnl: `I just picked to it against PHP 7.0.` Looks like this is missing the word `run`.

Also, you may as well run PHPCS against the highest PHP version available as that will give the best results if/when new tokens are introduced by PHP itself.

WordPressVIPMinimum/Sniffs/Actions/PreGetPostsSniff.php
```diff
@@ -40,7 +40,6 @@ class PreGetPostsSniff implements \PHP_CodeSniffer_Sniff {
 	 */
 	public function register() {
 		return Tokens::$functionNameTokens;
-
 	}//end register()
```
jrfnl: Still has an `// End` comment.

WordPressVIPMinimum/Sniffs/Cache/CacheValueOverrideSniff.php
```diff
@@ -38,7 +38,6 @@ class CacheValueOverrideSniff implements PHPCS_Sniff {
 	 */
 	public function register() {
 		return Tokens::$functionNameTokens;
-
 	}//end register()
```
jrfnl: Still has an `// End` comment.

WordPressVIPMinimum/Sniffs/Cache/CacheValueOverrideSniff.php
```diff
@@ -168,7 +166,6 @@ private function isVariableAssignment( $stackPtr ) {
 		}
 
 		return $previous;
-
 	}
 
 }//end class
```
jrfnl: Still has an `// End` comment.

WordPressVIPMinimum/Sniffs/Classes/DeclarationCompatibilitySniff.php
```diff
@@ -362,7 +362,6 @@ protected function loadFunctionNamesInScope( File $phpcsFile, $currScope ) {
 			$next                  = $phpcsFile->findNext( T_STRING, $i );
 			$this->_functionList[] = trim( $tokens[ $next ]['content'] );
 		}
-
 	}//end loadFunctionNamesInScope()
```
jrfnl: Still has an `// End` comment.

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -72,7 +72,6 @@ class IncludingFileSniff implements \PHP_CodeSniffer_Sniff {
 	 */
 	public function register() {
 		return Tokens::$includeTokens;
-
 	}//end register()
```
jrfnl: Still has an `// End` comment.

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -150,8 +149,7 @@ public function process( File $phpcsFile, $stackPtr ) {
 		} else {
 			$phpcsFile->addError( 'Absolute include path must be used. Use `get_template_directory()`, `get_stylesheet_directory()` or `plugin_dir_path()`.', $nextToken, 'IncludingFile' );
 			return;
-		}// End if().
-
+		}
 	}//end process()
```
jrfnl: Still has an `// End` comment.

composer.json
jrfnl: As long as there are no direct references to individual PHPCompatibility sniffs - i.e. inline PHPCS annotations to ignore a particular sniff or references in the ruleset -, the BC-break will be unnoticable.

AFAICS, for this project, some sniffs are excluded in the ruleset, so that will need updating when `9.0.0` comes out, but other than that, you should be fine.

jrfnl:
At @GaryJones' request I've looked into the failing build and have some more feedback on this PR/the VIP setup:

1. Travis: PHPCompatibility is being cloned for all builds, not just for the build in which it is used (`$SNIFF == 1`).
2. PHPCS does the test filtering *after* loading all tests for all standards (yeah... I know), this is problematic when you also have a clone of another standard installed which uses its own test framework - i.e. PHPCompatibility -. There is a solution available luckily - described here: https://github.com/PHPCompatibility/PHPCompatibility/blob/master/.github/CONTRIBUTING.md#issues-when-running-the-phpcs-unit-tests-for-another-standard
    Alternatively, install via Composer as that way the unit tests are not installed. (which is what I did for WPCS)
3.  Composer/Travis: if the unbounded WPCS dependency is correct, I would suggest adding a ruleset test and adding various versions of WPCS into the mix. I'm pretty sure those will fail as things stand now.
     This is also important because PHPCS is also mentioned as a dependency (which it is), so you'd want the PHPCS supported versions to be in sync with the PHPCS versions the WPCS version you support, supports.
    - Sniffs are included in the rulesets by name, if the sniff does not exist - either because it hasn't been added in WPCS yet when using an older version or if it has been renamed when using a newer version -, PHPCS will error out.
    - You are depending on WPCS for the PHPCS alias file and the unit test bootstraps, those were only introduced in one of the more WPCS versions. On that note: shouldn't the VIP standard have its own alias and test bootstrap file ?
4. If any of the VIP sniffs depend on WPCS - i.e. extend a WPCS sniff -, WPCS should be added to the Travis matrix anyway as you're now only testing against the latest WPCS release.

GaryJones:
> On that note: shouldn't the VIP standard have its own alias and test bootstrap file?

Yes, that would seem sensible.

> If any of the VIP sniffs depend on WPCS - i.e. extend a WPCS sniff

There are currently three sniffs that extend a WPCS sniff: [RegexpCompare](https://github.com/Automattic/VIP-Coding-Standards/blob/c74e8861cf3a0db3a3a1633cb9b7854144ec585b/WordPressVIPMinimum/Sniffs/VIP/RegexpCompareSniff.php), [RestrictedFunctions](https://github.com/Automattic/VIP-Coding-Standards/blob/c74e8861cf3a0db3a3a1633cb9b7854144ec585b/WordPressVIPMinimum/Sniffs/VIP/RestrictedFunctionsSniff.php), [RemoteRequestTimeout](https://github.com/Automattic/VIP-Coding-Standards/blob/c74e8861cf3a0db3a3a1633cb9b7854144ec585b/WordPressVIPMinimum/Sniffs/VIP/RemoteRequestTimeoutSniff.php).

I think a suggestion that may simplify things here, is to reset the supported minimums.

When WPCS 1.0.0 comes out, have that as the minimum.

WordPress.com VIP and Go both use PHP 7.2 is the minimum, so there doesn't seem to be much benefit in supporting earlier versions of PHP in this repo.

Is there any reason not to expect developers aiming for their code to run on VIP, to have the latest PHP_CodeSniffer (3.3.x) available? [WPCS is dropping 2.x](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1048) shortly anyway.

@david-binda @tomjn @sboisvert Thoughts?

tomjn:
For reference all VIP environments run PHP 7.2 at the moment, we have no environments running older code

tomjn:
There is some discussion as well on the plan for migrating VIP sniffs rom WPCS to this repo, @gudmdharalds has intentions on that front but is currently on holiday

tomjn:
> Is there any reason not to expect developers aiming for their code to run on VIP, to have the latest PHP_CodeSniffer (3.3.x) available? WPCS is dropping 2.x shortly anyway.

My preference would be to drop 2.x also

jrfnl:
> WordPress.com VIP and Go both use PHP 7.2 is the minimum, so there doesn't seem to be much benefit in supporting earlier versions of PHP in this repo.

1. As I noted earlier above, this repo isn't even tested against PHP 7.2.
2. While the wordpress.com platform may be running on PHP 7.2, this does not mean that all developers creating plugins/themes are on PHP 7.2. This is a dev tool, not a tool to run on the VIP platform.

So while you may choose to have a higher minimum PHP requirement than PHPCS itself, limiting it to PHP 7.2 or higher, would IMHO be a bit over the top and not something for which there is a valid reason.

> I think a suggestion that may simplify things here, is to reset the supported minimums.
> When WPCS 1.0.0 comes out, have that as the minimum.

At the very least set a bounded version requirement for WPCS, whether that requirement should be `^1.0.0` once it is out or something else, is another matter and as I stated above, depends largely on the ruleset configuration and what WPCS sniffs are being used.

Also, that doesn't invalidate my earlier remark about adding WPCS to the test matrix if more than one specific version of WPCS is supported.

> Is there any reason not to expect developers aiming for their code to run on VIP, to have the latest PHP_CodeSniffer (3.3.x) available? WPCS is dropping 2.x shortly anyway.

There is no reason for VIP not to drop support for PHPCS 2.x ahead of WPCS. As long as there is overlap between the PHPCS requirement, there is nothing stopping you there.

I've done the same for another standard which is based on WPCS a while back already.

composer.json
tomjn: +1

tomjn:
@GaryJones can you rebase on to `master` and resolve the conflicts?

GaryJones:
@tomjn Rebasing was going to be a conflict nightmare, so I branched from current `master`, cherry-picked _some_ of the commits from my `self-phpcs` branch, resolved the remaining conflicts, and then force pushed.

GaryJones:
Looking again at the Travis issue here.

jrfnl:
@GaryJones The solution for your Travis issue is described here: https://github.com/Automattic/VIP-Coding-Standards/pull/185#issuecomment-406283694

GaryJones:
@jrfnl Thanks - the Travis file needed bringing up to use a similar approach as to what we do at WPCS, so I refreshed it completely.

@tomjn Travis now passes :-)

- Tests against PHP 7.2 and PHP nightly.
- Tests against an old version of PHPCS (3.1.0) for all PHP versions except nightly. If at least `3.2.3` is [decided](https://github.com/Automattic/VIP-Coding-Standards/pull/198) to be the minimum version of PHPCS that is supported, then I think all of the fluff of explicitly using the PHPUnit 6 PHAR can be removed.
- Add cache directories for Composer.
- Use Composer to install dependencies, not git.
- Exclude low PHPCS version from running against PHP nightly (known failure of PHPCS < 3.3.1 with PHP nightly).
- Conditionally define constants.
- Disable Xdebug to speed up build time.
- Validate composer.json in before_install, so that any invalid JSON halts build immediately.
- Move composer install into `install` section.
- Use PHPUnit 6.5.11 for PHPCS 3.1.0, which doesn't support the PHPUnit 7 that gets installed with PHP 7.1+.
- Extract PHPUnit 6 version into constant, so future updates only have a single place to update this version number.
- Use shellcheck.net for recommended improvements to scripting.
- Add `xmllint --format` check for `WordPress-VIP-Go` ruleset file.

screenshot of command line showing phpcs and phpunit passing for VIPCS


As of this PR, VIPCS now passes it's own coding standards, and while we may still want to refine what those standards are (in the `.phpcs.xml.dist`), it's in a good place.

PR #202, opened 2018-08-09
PHPCS fixes

tomjn:
Inspired by #185 minus the changes to checks


tomjn:
@GaryJones about half of these are your fixes, it should make #185 much simpler/easier to review

PR #204, opened 2018-08-09
Fix error of WordPress.WP.GlobalVariables.OverrideProhibited not existing

rebeccahum:
```
ERROR: Referenced sniff "WordPress.WP.GlobalVariables.OverrideProhibited" does not exist
```
This should resolve it.


tomjn:
Or maybe I missed one, mistakes happen

PR #205, opened 2018-08-10
Add SVG files sniff for PHP openings

gudmdharalds:
Look for <?php and <?= in SVG files, and give a warning when that happens.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
+++ b/WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
@@ -39,6 +39,20 @@ class HTMLCodeSniff implements Sniff {
 	}
 
 	/**
+	 * Returns the file extension of a given file.
+	 *
+	 * @return string
+	 */
+	private function get_file_extension( $file_path ) {
+		return strtolower(
+			pathinfo(
+				$file_path,
+				PATHINFO_EXTENSION
+			)
+		);
+	}
+
+	/**
 	 * Processes the tokens that this sniff is interested in.
 	 *
 	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file where the token was found.
@@ -50,12 +64,9 @@ class HTMLCodeSniff implements Sniff {
 	public function process( File $phpcsFile, $stackPtr ) {
 		$tokens = $phpcsFile->getTokens();
 
-		// Make sure it is a SVG file.
-		$file_extension = strtolower(
-			pathinfo(
-				$phpcsFile->path,
-				PATHINFO_EXTENSION
-			)
+		// Get file extension
+		$file_extension = $this->get_file_extension(
+			$phpcsFile->path
 		);
 
 		// If not SVG file, ignore.
--- a/WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
+++ b/WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * WordPressVIPMinimum Coding Standard.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
+
+namespace WordPressVIPMinimum\Sniffs\SVG;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+
+/**
+ * This function generates an error when
+ * <?php or <? is found in a SVG file.
+ *
+ * The aim is to bring this to the notice of
+ * reviewers, so they can make sure the PHP-code
+ * is safe. The reason for this is that SVG files
+ * are often included in PHP files, but manually
+ * scanning SVG files takes alot of time and is prone
+ * to errors.
+ */
+class HTMLCodeSniff implements Sniff {
+	public $supportedTokenizers = array(
+		'PHP',
+	);
+
+	/**
+	 * Returns the token types that this sniff is interested in.
+	 *
+	 * We want everything function-related.
+	 *
+	 * @return array(int)
+	 */
+	public function register() {
+		return array(
+			T_OPEN_TAG,
+		);
+	}
+
+	/**
+	 * Processes the tokens that this sniff is interested in.
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file where the token was found.
+	 * @param int						 $stackPtr  The position in the stack where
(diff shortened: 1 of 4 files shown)


WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
```diff
@@ -0,0 +1,82 @@
...
+	 *                                               the token was found.
+	 *
+	 * @return void
+	 */
+	public function process( File $phpcsFile, $stackPtr ) {
+		// Get tokens.
+		$tokens = $phpcsFile->getTokens();
+
+		$nxt = $phpcsFile->findNext(
+			T_OPEN_TAG,
```
david-binda: the `process` method is entered only when `T_OPEN_TAG` is encountered. That say, mere entering of this method makes it explicit that an open tag has been found, and there is no need to check whether it really is an open tag, or of what type.

Thus, in case we are in this method, we can directly return the Warning.

WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
```diff
@@ -0,0 +1,82 @@
...
+	/**
+	 * Returns the token types that this sniff is interested in.
+	 *
+	 * We want everything function-related.
+	 *
+	 * @return array(int)
+	 */
+	public function register() {
+		return array(
+			T_OPEN_TAG,
```
david-binda: There is also `T_OPEN_TAG_WITH_ECHO` checking for `<?=` which should be part of this sniff, imho.

WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
```diff
@@ -0,0 +1,82 @@
...
+		);
+
+		$tokens[ $nxt ]['content'] =
+			strtolower( $tokens[ $nxt ]['content'] );
+
+		$found1 =
+			strpos( $tokens[ $nxt ]['content'], '<?php' );
+
+		$found2 =
+			strpos( $tokens[ $nxt ]['content'], '<?' );
```
david-binda: Unless short open tags are allowed in the PHP config on the server where this runs, those won't be catched by `T_OPEN_TAG`.

Only way to catch short open tags is to have short open tags allowed on the server. But it only makes sense in case it mirrors the production config - I believe that short open tags are disallowed on both, WPCOM and VIP GO now.

Here's a related issue in PHPCS covering the topic of short open tags and it's catching: https://github.com/squizlabs/PHP_CodeSniffer/issues/1571 and here's a direct link to how it's handled in the Generic.PHP.DisallowShortOpenTag sniff: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Generic/Sniffs/PHP/DisallowShortOpenTagSniff.php#L32,L34

WordPressVIPMinimum/Tests/SVG/HTMLCodeSniff.php
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	/**
+	 * Returns the lines where warnings should occur.
+	 *
+	 * @return array <int line number> => <int number of warnings>
+	 */
+	public function getWarningList() {
+		return array(
+			7 => 1,
```
GaryJones: Where's the test file that this relates to? I don't see it in this PR.

gudmdharalds: Has been added now. :-)

WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
gudmdharalds: I think we will let it suffice to look for `<?php` and `<?=` because of how the configuration is on our servers.

gudmdharalds: And, to add: this test will be in conjunction with the `Generic.PHP.DisallowShortOpenTag` sniff.

WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
```diff
@@ -0,0 +1,72 @@
...
+	 * @param int                         $stackPtr  The position in the stack where
+	 *                                               the token was found.
+	 *
+	 * @return void
+	 */
+	public function process( File $phpcsFile, $stackPtr ) {
+		$tokens = $phpcsFile->getTokens();
+
+		// Make sure it is a SVG file.
+		$file_extension = strtolower(
```
GaryJones: This `strtolower( ... )` could be extracted into its own method to make the code more self-documenting. Line 62 would then be something like:

```php
if ( 'svg' !== $this->get_file_extension( $phpcsFile->path ) ) {
```

WordPressVIPMinimum/Tests/SVG/HTMLCodeSniff.svg
```diff
@@ -0,0 +1,26 @@
+<svg width="299px" height="31px" viewBox="0 0 299 11" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+	<!-- Generator: Tezt -->
+	<title>Page 100</title>
+	<desc>Created with Sketch.</desc>
+
+
+
+<?php
```
GaryJones: I'd like to see some single-line opening and closing tags, and with something in between them, to ensure that the logic handles some likely uses. Just having the opening and closing tags (both types), nicely at the start of different lines, with nothing in between, isn't pushing to find all the edge cases.

gudmdharalds: Thanks! I've added this now :-)

WordPressVIPMinimum/Sniffs/SVG/HTMLCodeSniff.php
gudmdharalds: Done :-)

WordPressVIPMinimum/Tests/SVG/HTMLCodeSniff.svg
GaryJones: Sorry - I wasn't clear - I wanted to see multiple efforts with and without extra content:

```
<?php
?><?php echo 'foo'; ?><?= ?><?php ?><?=

?>
```

etc.

PR #212, opened 2018-08-29
Increases the PHPCS and PHP minimum versions

tomjn:
Increases them to supported and known good values

Specifically PHP 5.6 and PHPCS v3.2.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,59 +3,18 @@ sudo: false
 dist: trusty
 
 cache:
-  apt: true
   directories:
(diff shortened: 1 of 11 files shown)


tomjn:
Closes #212

GaryJones:
May need to remove the 5.4 and 5.5 references from `.travis.yml` to make this pass without having to re-start each of the jobs.

https://github.com/Automattic/VIP-Coding-Standards/pull/210 will then update it further anyway.

PR #213, opened 2018-08-30
Remove AssignmentinCondition rule

rebeccahum:
`Variable assignment found within a condition. Did you mean to do a comparison?` 

I find that this flags a false positive 99.9999% of the time since it's more of a...coding style?  Would be nice to remove altogether for noise purposes on Go.


rebeccahum:
plz ignore my git incapabilities

PR #215, opened 2018-08-30
Downgrade severity of Variable assignment condition rule

rebeccahum:
`Variable assignment found within a condition. Did you mean to do a comparison?`

I find that this flags a false positive 99.9999% of the time since it's more of a...coding style? Would be nice to remove altogether for noise purposes on Go.

### DIFF
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -233,6 +233,10 @@
 		<type>warning</type>
 		<severity>3</severity>
 	</rule>
+	<rule ref="WordPress.CodeAnalysis.AssignmentInCondition.Found">
+		<type>warning</type>
+		<severity>1</severity>
+	</rule>
 	<rule ref="WordPressVIPMinimum.Files.IncludingFile.IncludingFile">
 		<type>warning</type>
 		<severity>3</severity>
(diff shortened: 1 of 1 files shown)


GaryJones:
My comment from Slack:

> It is a style, but one that could lead to hard-to-find bugs.  If one is assigning to a variable, then the code is generally going to be easier to read if it's done as a normal assignment statement first, and _then_ use that variable in the conditional. It removes the ambiguity for whether it was meant to be an assignment or comparison. For WPCS at least, the presence of that check is a good reason that we're considering scrapping Yoda conditions...

Then @david-binda said:

> That’s also what happened at VIP some time ago. We replaced Yoda by Variable assignment within a condition…

GaryJones:
> Perhaps we can move it to a warning severity of 1 and for SRs, silence anything that's under 2?

In terms of being able to efficiently conduct code reviews, which benefits clients in many ways, I'd expect A8C to want to require code be as unambiguous as possible.

sboisvert:
>I'd expect A8C to want to require code be as unambiguous as possible.

We've previously been all about requiring stuff and we're moving away from it. We want to recommend the best practices without getting in the way of people. So instead of requiring people to change their code we want to have documentation that helps folks understand "why" we recommend this. What are the benefits of it.

I think this is safe to flag as a 1 or a 3 in terms of severity. 

>and for SRs, silence anything that's under 2? @sboisvert thoughts?

I think that depends on what you mean as "for SR". I think in the near term we're going to remove all human involvement in the SR which means we don't need to "silence" things as much as let clients pick the level they want. In the short term when we're going a human review of the SR we should only look at things that are > 5 so we don't need to add an additional silencing for things below 3 :)

PR #217, opened 2018-09-05
Travis: Remove PHPUnit 6 workaround

GaryJones:
PHPUnit 6 was needed when the lowest supported/tested version of PHPCS was 3.1.0.

PHPCS 3.2.3 (which is the currently lowest supported version) claims to support PHPUnit 7, so we can let the Composer constraint of `"phpunit/phpunit": "*"` dictate which version of PHPUnit is installed and used when running the test, depending on the version of PHP being used.

Currently:

| PHP Version 	| PHPUnit Version 	|
|-------------	|-----------------	|
|     5.6     	|      5.7.27     	|
|     7.0     	|      6.5.12     	|
|     7.1     	|      7.3.4      	|
|     7.2     	|      7.3.4      	|
|   Nightly   	|      7.3.4      	|

Also adds explicit stage naming so that we avoid a situation like de3061d654cc4c46bc6a79ad996b2d947312dff1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,28 +15,13 @@ before_install:
   # https://twitter.com/kelunik/status/954242454676475904
   - phpenv config-rm xdebug.ini || echo 'No xdebug config.'
 
-  # Handle new PHP version with old PHPCS version. Version of PHPUnit 6 is the current latest release.
-  - |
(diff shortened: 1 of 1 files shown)


GaryJones:
Turns out I was looking at the wrong Travis results on my own fork. I'll get this failure addressed.

Issue #20, opened 2017-01-27
Flag SERVER vars


joshbetz:
* `HTTP_X_IP_TRAIL`
* `HTTP_X_FORWARDED_FOR`
* `REMOTE_ADDR`

rebeccahum:
Added as per #219.

PR #219, opened 2018-09-12
Fix detection of double quotes, add additional server variables and u…

rebeccahum:
As per #20, this PR adds the additional server variables and updates the related unit tests.

In addition, it also detects the usage of double quotes.

E.g.
```
echo $_SERVER[ "REMOTE_ADDR" ];
```

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Variables/ServerVariablesSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Variables/ServerVariablesSniff.php
@@ -24,10 +24,15 @@ class ServerVariablesSniff implements Sniff {
 	 * @var array
 	 */
 	public $restrictedVariables = array(
-		'PHP_AUTH_PW',
-		'HTTP_X_IP_TRAIL',
-		'HTTP_X_FORWARDED_FOR',
-		'REMOTE_ADDR',
+		'authVariables' => array( 
+			'PHP_AUTH_USER' => true,
+			'PHP_AUTH_PW'   => true 
+		),
+		'userControlledVariables' => array( 
+			'HTTP_X_IP_TRAIL' 	   => true,
+			'HTTP_X_FORWARDED_FOR' => true,
+			'REMOTE_ADDR' 		   => true, 
+		)
 	);
 
 	/**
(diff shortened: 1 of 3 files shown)


tomjn:
Looks like the unit tests expectations need adjusting to account for the new errors being thrown

rebeccahum:
@tomjn My apologies, I've fixed it so it should be OK now.

WordPressVIPMinimum/Sniffs/Variables/ServerVariablesSniff.php
```diff
@@ -56,14 +59,22 @@ public function process( File $phpcsFile, $stackPtr ) {
...
 
 		if ( false === in_array( $variableName, $this->restrictedVariables, true ) ) {
 			// Not the variable we are looking for.
 			return;
 		}
 
-		$phpcsFile->addError( 'Basic authentication should not be handled via PHP code.', $stackPtr, 'ServerVariables' );
+		if ( 'PHP_AUTH_PW' === $variableName ) {
+			$phpcsFile->addError( 'Basic authentication should not be handled via PHP code.', $stackPtr, 'ServerVariables' );
+		} elseif ( 'HTTP_X_IP_TRAIL' === $variableName || 'HTTP_X_FORWARDED_FOR' === $variableName || 'REMOTE_ADDR' === $variableName ) {
```
GaryJones: Rather than comparing `=== $variableName` three times (and potentially other strings in the future), why not add the current three items to an array (use associative array with strings as keys and `true` as value - see [here](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/1bc614137d5ed913ab2df00208b1edf5279dcb8e/WordPress/Sniff.php#L25-L34) for reason), and then use `isset()`?

```php
/**
 * {Add docs here as to why these are special}
 *
 * @since ...
 */
private $user_controlled_variables = array(
    'HTTP_X_IP_TRAIL'      => true,
    'HTTP_X_FORWARDED_FOR' => true,
    'REMOTE_ADDR'          => true,
);

...

} elseif ( isset( $this->user_controlled_variables[ $variableName ] ) {
```

The data (property) is then separate from the logic (method), the data can be documented (property DocBlock), and the logic is more self-documenting (property name).

GaryJones:
@rebeccahum Travis is reporting some WPCS issues that should be addressed. Looks like some tabs are mixed in with spaces in a few places.

PR #227, opened 2018-10-10
Docs: Update VIP link references

GaryJones:
The WordPress.com VIP team moved the coding standards documentation around, since they now have different documents for their different platforms:

- [WPCOM platform](https://lobby.vip.wordpress.com/wordpress-com-documentation/code-review-what-we-look-for/) (only available to those with VIP Lobby access)
- [VIP-Go platform](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)

This updates the out of date references for links to the Lobby.

See https://github.com/Automattic/vip-issues/issues/160

### DIFF
--- a/WordPressVIPMinimum/ruleset.xml
+++ b/WordPressVIPMinimum/ruleset.xml
@@ -2,6 +2,10 @@
 <ruleset name="WordPress VIP Minimum">
 	<description>WordPress VIP Minimum Coding Standards</description>
 
+	<!-- Link references are platform-specific. See
+		https://vip.wordpress.com/documentation/vip-go/understanding-codebase-differences/
+		for the differences between the platforms -->
+
 	<autoload>./PHPCSCompatibility.php</autoload>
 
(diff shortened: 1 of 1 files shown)


GaryJones:
Fair point. Though for anyone _just_ checking `WordPressVIPMinimum`, it may seem odd to be reading the VIP-Go Docs, which may even check things differently (different severity levels, warning instead of error, etc.)

How about a comment at the top, and then adding both links, where relevant?

```xml
<!-- Link references are platform-specific. See 
https://vip.wordpress.com/documentation/vip-go/understanding-codebase-differences/ 
for the differences between the platforms -->

...

<!-- WordPress.com: https://lobby.vip.wordpress.com/wordpress-com-documentation/code-review-what-we-look-for/#settings-alteration -->
<!-- VIP-Go: https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/#settings-alteration -->

rebeccahum:
Yup, that makes more sense.

rebeccahum:
Created issue #235.

Issue #87, opened 2017-06-22
Adding  'upload_mimes' filter should generate a warning 


gudmdharalds:
For instance the following should generate a warning:

```
function abc_upload_mime_types( $mime_types = array() ) {
   $mime_types['test'] = 'text/testing123';
   return $mime_types;
}

add_filter( 'upload_mimes', 'abc_upload_mime_types' );
```

Issue #187, opened 2018-07-25
WPCS 2.0: Migrate Deprecated VIP Sniffs


tomjn:
WPCS 1.0.0 deprecates VIP sniffs, with the intention that they be removed in a later release. We should copy these sniffs to VIP Minimum appropriatley so that our own sniffs and rules aren't broken by a future release

GaryJones:
All but one of the `WordPressVIPMinimum/Sniffs` implement a core PHPCS sniff (`PHP_CodeSniffer_Sniff` currently, but will be `PHP_CodeSniffer\Sniffs\Sniff` if #190 happens first). There is no dependency on WPCS here.

The other Sniff, `DeclarationCompatibilitySniff`, extends `\PHP_CodeSniffer_Standards_AbstractScopeSniff`. There is no dependency on WPCS here.

----

The `WordPressVIPMinimum` [`ruleset.xml`](https://github.com/Automattic/VIP-Coding-Standards/blob/dae432a40e20ccc4f2122c9b950a6ab510d4ec81/WordPressVIPMinimum/ruleset.xml) includes the following `WordPress.VIP` references:

```xml
	<rule ref="WordPress.VIP">
		<exclude name="WordPress.VIP.ValidatedSanitizedInput.MissingUnslash"/>
		<exclude name="WordPress.VIP.RestrictedFunctions"/>
	</rule>
...
	<rule ref="WordPress.VIP.AdminBarRemoval.DeprecatedSniff">
		<severity>0</severity>
	</rule>
	<rule ref="WordPress.VIP.CronInterval.DeprecatedSniff">
		<severity>0</severity>
	</rule>
	<rule ref="WordPress.VIP.DirectDatabaseQuery.DeprecatedSniff">
		<severity>0</severity>
	</rule>
	<rule ref="WordPress.VIP.FileSystemWritesDisallow.DeprecatedSniff">
		<severity>0</severity>
	</rule>
	<rule ref="WordPress.VIP.OrderByRand.DeprecatedSniff">
		<severity>0</severity>
	</rule>
	<rule ref="WordPress.VIP.PluginMenuSlug.DeprecatedSniff">
		<severity>0</severity>
	</rule>
	<rule ref="WordPressVIPMinimum.VIP.RestrictedFunctions.DeprecatedSniff">
		<severity>0</severity>
	</rule>
	<rule ref="WordPress.VIP.RestrictedVariables.DeprecatedSniff">
		<severity>0</severity>
[...]

PR #231, opened 2018-10-16
Add a sniff for "Restricted Filters"

rebeccahum:
As per #87, I think we should add a sniff that will detect a list of restricted filters...with `upload_mimes` being the first one on it!

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * WordPressVIPMinimum Coding Standard.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
+
+namespace WordPressVIPMinimum\Sniffs\Filters;
+
+use PHP_CodeSniffer_File as File;
+use PHP_CodeSniffer_Tokens as Tokens;
+
+/**
+ * This sniff restricts usage of some filters
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
+class RestrictedFilterSniff implements \PHP_CodeSniffer_Sniff {
+
+	/**
+	 * List of restricted filter names.
+	 *
+	 * @var array
+	 */
+	public $restrictedFilters = array(
+		'upload_mimes',
+	);
+
+	/**
+	 * Returns the token types that this sniff is interested in.
+	 *
+	 * @return array(int)
+	 */
+	public function register() {
+		return Tokens::$functionNameTokens;
+	}
+
+	/**
+	 * Processes the tokens that this sniff is interested in.
+	 *
+	 * @param \PHP_CodeSniffer\Files\File $phpcsFile The file where the token was found.
+	 * @param int                         $stackPtr  The position in the stack where
+	 *                                               the token was found.
+	 *
+	 * @return void
+	 */
+	public function process( File $phpcsFile, $stackPtr ) {
+
+		$tokens = $phpcsFile->getTokens();
+
+		$functionName = $tokens[ $stackPtr ]['content'];
+
+		if ( 'add_filter' !== $functionName ) {
+			return;
+		}
+
+		$filterNamePtr = $phpcsFile->findNext(
+			array_merge( Tokens::$emptyTokens, array( T_OPEN_PARENTHESIS ) ), // types.
+			$stackPtr + 1, // start.
+			null, // end.
+			true, // exclude.
+			null, // value.
+			true // local.
+		);
+
+		$filterName = str_replace( array( "'", '"' ), '', $tokens[ $filterNamePtr ]['content'] );
+
+
+		if ( false === in_array( $filterName, $this->restrictedFilters, true ) ) {
+			// Not the filter we are looking for.
+			return;
+		}
+
+		if ( 'upload_mimes' === $filterName ) {
+			$phpcsFile->addWarning( 'Please ensure that the mime types being filtered do not include SVG. Manual inspection required.', $stackPtr, 'RestrictedFilter' );
+		}
+
+	}
+
+}
--- a/WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
@@ -23,7 +23,7 @@ class RestrictedFilterSniff implements \PHP_CodeSniffer_Sniff {
 	 * @var array
 	 */
 	public $restrictedFilters = array(
-		'upload_mimes',
+		'upload_mimes' => true,
 	);
 
 	/**
@@ -65,14 +65,8 @@ class RestrictedFilterSniff implements \PHP_CodeSniffer_Sniff {
 
 		$filterName = str_replace( array( "'", '"' ), '', $tokens[ $filterNamePtr ]['content'] );
 
-
-		if ( false === in_array( $filterName, $this->restrictedFilters, true ) ) {
-			// Not the filter we are looking for.
-			return;
-		}
-
-		if ( 'upload_mimes' === $filterName ) {
-			$phpcsFile->addWarning( 'Please ensure that the mime types being filtered do not include SVG. Manual inspection required.', $stackPtr, 'RestrictedFilter' );
+		if ( isset( $this->restrictedFilters[ $filterName ] ) && 'upload_mimes' === $filterName ) {
+			$phpcsFile->addWarning( 'Please ensure that the mimes being filtered do not include insecure types (e.g. SVG). Manual inspection required.', $stackPtr, 'UploadMimesRestrictedFilter' );
 		}
 
 	}
--- a/WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.inc
+++ b/WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.inc
@@ -1,5 +1,5 @@
 <?php
 
-add_filter( 'test_filter', 'good_example_function' );
+add_filter( 'test_filter', 'good_example_function' ); 	// this one shouldn't trigger anything
 
 add_filter( 'upload_mimes', 'bad_example_function' );
--- a/WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.inc
+++ b/WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.inc
@@ -1,0 +1,5 @@
+<?php
+
+add_filter( 'test_filter', 'good_example_function' );
+
+add_filter( 'upload_mimes', 'bad_example_function' );
--- a/WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.php
+++ b/WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * Unit test class for WordPressVIPMinimum Coding Standard.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
+
+namespace WordPressVIPMinimum\Tests\Filters;
+
+use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
+/**
+ * Unit test class for the Filters/AlwaysReturn sniff.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
+class RestrictedFilterUnitTest extends AbstractSniffUnitTest {
+
+	/**
+	 * Returns the lines where errors should occur.
+	 *
+	 * @return array <int line number> => <int number of errors>
+	 */
+	public function getErrorList() {
+		return array();
+	}
+
+	/**
+	 * Returns the lines where warnings should occur.
+	 *
+	 * @return array <int line number> => <int number of warnings>
+	 */
+	public function getWarningList() {
+		return array(
+			5 => 1,
+		);
+	}
+
+}


WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.inc
```diff
@@ -0,0 +1,5 @@
+<?php
+
+add_filter( 'test_filter', 'good_example_function' ); 	// this one shouldn't trigger anything
+
+add_filter( 'upload_mimes', 'bad_example_function' );
```
GaryJones: I'd like to see more tests please:

 - combinations of whitespace / no-whitespace before and after the filter name arg i.e. `add_filter('upload_mimes' , 'foo' );`
 - string with upper case characters (`'UPLoad_MimEs'`)
 - string with double quotes (`"upload_mimes"`)
 - strings with concatenation (`'upload_' . 'mimes'`) 
 - strings with compounded strings/vars
   - `$x = 'd_m'; add_filter( 'uploa' . $x . 'imes', 'foo')`
   - `$x = 'd_m'; add_filter( "uploa{$x}imes", 'foo')`
 - `add_filter()` calls with anonymous function callbacks or other callables.
 - `add_filter()` called dynamically (may be overkill?)
   - `$x = 'add_filter'; $x( 'upload_mimes', 'foo' );`
 - etc.

See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Tests/NamingConventions/ValidHookNameUnitTest.1.inc and sibling files for other ideas of ways folks could try and bypass the check by mangling the filter name string.

Your code may already cope with some of those (like double-quotes), but adding them as tests just proves that for the future.

Also, keep a clear linespace on the last line (adjust your editor to do this, since the repo doesn't yet have an `.editorconfig` file).

WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.inc
```diff
@@ -0,0 +1,5 @@
+<?php
+
+add_filter( 'test_filter', 'good_example_function' ); 	// this one shouldn't trigger anything
```
GaryJones: Just an `// Ok.` is fine here as a comment.

Don't use tabs to separate the comment from the code - a single space is fine.

Remember full-stop, since that's needed as part of `WordPress-Docs`

WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.php
```diff
@@ -0,0 +1,38 @@
...
+ * Unit test class for WordPressVIPMinimum Coding Standard.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
+
+namespace WordPressVIPMinimum\Tests\Filters;
+
+use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
+/**
+ * Unit test class for the Filters/AlwaysReturn sniff.
```
GaryJones: Summary needs updating.

WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.php
```diff
@@ -0,0 +1,38 @@
...
+ *
```
GaryJones: Add a `@since` version number for whatever the next version will be (presumably `0.4.0`).

WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+namespace WordPressVIPMinimum\Sniffs\Filters;
+
+use PHP_CodeSniffer_File as File;
+use PHP_CodeSniffer_Tokens as Tokens;
+
+/**
+ * This sniff restricts usage of some filters
+ *
```
GaryJones: Add a `@since` version number for whatever the next version will be (presumably `0.4.0`).

WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
```diff
@@ -0,0 +1,72 @@
...
+ * @package VIPCS\WordPressVIPMinimum
+ */
+
+namespace WordPressVIPMinimum\Sniffs\Filters;
+
+use PHP_CodeSniffer_File as File;
+use PHP_CodeSniffer_Tokens as Tokens;
+
+/**
+ * This sniff restricts usage of some filters
```
GaryJones: Summaries for classes need trailing full stops.

WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
```diff
@@ -0,0 +1,72 @@
...
+	 * @var array
+	 */
+	public $restrictedFilters = array(
+		'upload_mimes' => true,
+	);
+
+	/**
+	 * Returns the token types that this sniff is interested in.
+	 *
+	 * @return array(int)
```
GaryJones: It's fine to follow what [PHPCS does](https://github.com/squizlabs/PHP_CodeSniffer/blob/c056352b9c27a951775b526c01bf1223364df558/src/Util/Tokens.php#L550) for the type here: `array<int, int>`.

[Generics notation is supported in the draft PSR-5](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#collections).

GaryJones:
Also, whilst the PR creates the sniff, the test file, and the test inc file, it doesn't actually add the sniff to any ruleset, along with a link reference to the relevant documentation page(s).

WordPress-VIP-Go/ruleset.xml
```diff
@@ -104,6 +104,10 @@
 
 
 	<!-- Warnings and other things -->
+	<rule ref="WordPressVIPMinimum.Filters.RestrictedFilter.UploadMimes">
+		<type>warning</type>
+		<severity>10</severity>
```
GaryJones: Why the choice of `10` here? Not saying it's wrong, just looking for a rationale between the items that are `7` and `10` here.

WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
```diff
@@ -55,18 +57,59 @@ public function process( File $phpcsFile, $stackPtr ) {
...
+			null,
+			false,
+			null, 
+			true
 		);
 
-		$filterName = str_replace( array( "'", '"' ), '', $tokens[ $filterNamePtr ]['content'] );
+		$filterName = $this->transformString( $tokens[ $filterNamePtr ]['content'] );
+
+		// if concatenation is found, build $filterName.
```
GaryJones: Capital needed for the start of comments, per `WordPress-Docs`.

WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
```diff
@@ -55,18 +57,59 @@ public function process( File $phpcsFile, $stackPtr ) {
...
+
+	/**
+	 * Transform string.
+	 *
+	 * @param string $string The name of the filter.
+	 *
+	 * @return string
+	 */
+	private function transformString( $string ) {
+		$string = str_replace( array( "'", '"' ), '', $string ); // remove quotes (double and single).
```
GaryJones: Capital needed for the start of comments, per `WordPress-Docs`.

Code base is [PHP 5.6+](https://github.com/Automattic/VIP-Coding-Standards/blob/a1e3be751a7edd8678965a504a786130db7ef293/composer.json#L13), so short array syntax can be used.

WordPressVIPMinimum/Sniffs/Filters/RestrictedFilterSniff.php
```diff
@@ -55,18 +57,59 @@ public function process( File $phpcsFile, $stackPtr ) {
...
+		$string = strtolower( $string ); // make sure we don't have weird caps.
```
GaryJones: Capital needed for the start of comments, per `WordPress-Docs`.

WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.inc
```diff
@@ -1,5 +1,15 @@
 <?php
 
-add_filter( 'test_filter', 'good_example_function' ); 	// this one shouldn't trigger anything
+add_filter( 'test_filter', 'good_example_function' ); // Ok.
 
-add_filter( 'upload_mimes', 'bad_example_function' );
\ No newline at end of file
+// warnings.
+add_filter( 'upload_mimes', 'bad_example_function' );
```
GaryJones: Add a short comment after each one, explaining what it is testing:

```php
add_filter( 'upload_mimes', 'bad_example_function' ); // Simple string.
add_filter('upload_mimes' ,'bad_example_function'); // Incorrect spacing.
add_filter(  'upload_mimes','bad_example_function'); // Incorrect spacing.
add_filter( "upload_mimes" ,'bad_example_function'); // Double quotes.
add_filter( 'upLoad_mimeS' ,'bad_example_function'); // Uppercase characters.
add_filter( 'upload_' . 'mimes' ,'bad_example_function'); // Single concatenation.
add_filter( 'upl' . 'oad_' . 'mimes' ,'bad_example_function'); // Multiple concatenation.
add_filter( 'upload_mimes', function() { // Anonymous callback.
	// Do stuff.	
});

File: WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.inc [hunk 13]
```diff
@@ -1,5 +1,15 @@
...
+add_filter('upload_mimes' ,'bad_example_function');
+add_filter(  'upload_mimes','bad_example_function');
+add_filter( "upload_mimes" ,'bad_example_function');
+add_filter( 'upLoad_mimeS' ,'bad_example_function');
+add_filter( 'upload_' . 'mimes' ,'bad_example_function');
+add_filter( 'upl' . 'oad_' . 'mimes' ,'bad_example_function');
+add_filter( 'upload_mimes', function() {
+	// do stuff.	
+});
```
GaryJones: Not seeing any tests for dynamic hook names?

I guess someone would definitely be actively trying to game the system if they were using this, so it could always be added later if necessary.

GaryJones:
I've spoken to @jrfnl to see if PHPCS or WPCS has some existing battle-tested code for getting the first/second/third/etc argument from a function call, that might already exist and cater for the possibility of dynamic string arguments.

She's not got back to me yet, but [`get_function_call_parameters()`](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/1cb6a74615be51390776c3c22438c21b010c1737/WordPress/Sniff.php#L2125-L2214) looks to be what I'm after, and this is used in a  [AbstractFunctionParameterSniff](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/AbstractFunctionParameterSniff.php) which is what [some sniffs](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/search?q=AbstractFunctionParameterSniff&unscoped_q=AbstractFunctionParameterSniff) extend. This allows simply checking for `$parameters[1]`, for instance.

WPCS has [this Sniff](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniff.php), which apart from containing a bunch of properties, also contains some re-usable methods. VIPCS doesn't yet have anything like this. Since we can use PHP 5.6, we could separate this into Traits that sniffs like the one in this PR could then use.

This would be a separate Issue, but I think one that would be worthwhile to do first, and then make use of it for a PR like this one. What do you think @rebeccahum 	?

WordPressVIPMinimum/Tests/Filters/RestrictedFilterUnitTest.inc
GaryJones: Actually, since these are tokens, it may not be possible parse the variable value and see if variable + string = upload_mimes.

GaryJones:
@rebeccahum So, give this a try, as a straight replacement for the Sniff:

```php
<?php
/**
 * WordPressVIPMinimum Coding Standard.
 *
 */

namespace WordPressVIPMinimum\Sniffs\Filters;

use WordPress\AbstractFunctionParameterSniff;

/**
 * This sniff restricts usage of some filters.
 *
 *
 * @since 0.4.0
 */
class RestrictedFilterSniff extends AbstractFunctionParameterSniff {

	/**
	 * The group name for this group of functions.
	 *
	 * @var string
	 */
	protected $group_name = 'restricted_filters';

	/**
	 * Functions this sniff is looking for.
	 *
	 * @var array The only requirement for this array is that the top level
	 *            array keys are the names of the functions you're looking for.
	 *            Other than that, the array can have arbitrary content
	 *            depending on your needs.
	 */
	protected $target_functions = [
		'add_filter' => true,
		'add_action' => true,
	];

	/**
	 * List of restricted filter names.
	 *
	 * @var array
	 */
	private $restricted_filters = [
		'upload_mimes' => [
			'error'     => 'Please ensure that the mimes being filtered do not include insecure types (e.g. SVG). Manual inspection required.',
			'errorcode' => 'UploadMimes',
		],
	];

	/**
	 * Process the parameters of a matched function.
	 *
	 * @param int    $stackPtr        The position of the current token in the stack.
	 * @param array  $group_name      The name of the group which was matched.
	 * @param string $matched_content The token content (function name) which was matched.
	 * @param array  $parameters      Array with information about the parameters.
	 * @return int|void Integer stack pointer to skip forward or void to continue
	 *                  normal file processing.
	 */
	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
		foreach ( $this->restricted_filters as $restricted_filter => $filter_args ) {
			if ( $this->normalize_filter_name( $parameters[1] ) === $restricted_filter ) {
				$this->phpcsFile->addWarning( $filter_args['error'], $stackPtr, $filter_args['errorcode'] );
			}
		}
	}

	/**
	 * Transform string.
	 *
	 * @param array $parameter Array with information about a parameter.
	 * @return string
	 */
	private function normalize_filter_name( $parameter ) {
		// If concatenation is found, build filter name.
		$concat_ptr = $this->phpcsFile->findNext(
			T_STRING_CONCAT,
			$parameter['start'],
			$parameter['end'],
			false,
			null,
			true
		);

		if ( $concat_ptr ) {
			$filter_name = '';
			for ( $i = $parameter['start'] + 1; $i < $parameter['end']; $i++ ) {
				if ( T_CONSTANT_ENCAPSED_STRING === $this->tokens[ $i ]['code'] ) {
					$filter_name .= str_replace( [ "'", '"' ], '', $this->tokens[ $i ]['content'] );
				}
			}
		} else {
			$filter_name = $parameter['raw'];
		}

		// Remove quotes (double and single), and use lowercase.
		return trim( strtolower( str_replace( [ "'", '"' ], '', $filter_name ) ) );
	}
}

```

All of the existing unit tests pass, and there's a couple of others as well:

```php
add_filter( "upload_" . 'mimes' ,'bad_example_function'); // Single concatenation with double and single quotes.
add_filter( 'upl' . "oad_" . "mimes" ,'bad_example_function'); // Multiple concatenation with double and single quotes.
add_action( 'upload_mimes', 'bad_example_function' ); // Check `add_action()`, which is an alias for `add_filter()`.
```

The `normalize_filter_name()` could be moved to a base VIPCS sniff class if it was needed to be re-used, and the concat logic could be extracted into its own `concatenate_parameter()` method or something.

We've used the logic from WPCS, which makes it easier to generalise things and make this sniff more re-usable. Your existing warning message referred directly to the upload_mimes filter, for instance, which means we couldn't add different warning messages for other filters. By moving having the config array of `error` and `errorcode`, this is solved.

I've got a local branch with this on, which cherry-picked the commits from the branch on your repo, so I can wrap this up with a fresh PR if you wish - you'd still get author credits for your commits, even after squashing :-)

rebeccahum:
Thank you so much, this looks great.  This makes sense to me since we can re-use the logic from WPCS.  However, are there any downsides on having WPCS dependencies?

rebeccahum:
Thank you so much, this looks great. This makes sense to me since we can re-use the logic from WPCS. However, are there any downsides on having WPCS dependencies?

rebeccahum:
@GaryJones Thank you so much, this looks great. This makes sense to me since we can re-use the logic from WPCS. However, are there any downsides on having WPCS dependencies?

GaryJones:
> However, are there any downsides on having WPCS dependencies?

We already _had_ a `require` dependency for WPCS, since right now, a few sniffs depend on sniffs from `WordPress-VIP`. See #187.

I think that #187 could still go ahead, since the code that _this_ PR relies on is not a concrete sniff, but an abstract one that provides re-usable logic, not specific implementation for the end problem.

However, of more interest is https://github.com/squizlabs/PHP_CodeSniffer/issues/2189 which, if implemented, gives a definite path to not have to rely on WPCS for the function parameter logic.

GaryJones:
Approved, but since I've not officially started at A8C yet, it may be worth getting another pair of eyes. cc @sboisvert @tomjn @gudmdharalds.

Issue #81, opened 2017-05-08
Flag wp_cache_set() with very short expiry-time


gudmdharalds:
We should flag any `wp_cache_set()` calls with very low expiry-times with a 'warning', because of their ineffectiveness.  

For instance, this is not flagged:

```
wp_cache_set( 
    'test-key',
    array(
        'k' => 'v',
    ),
    'test-group',
    60
);
```

I suggest we flag everything that is cached for less than 10 minutes, as that period should be acceptable for a range of data.

david-binda:
> I suggest we flag everything that is cached for less than 10 minutes, as that period should be acceptable for a range of data.

We do quite often suggest clients to use 300 / 5mins for caching data which need to be "fresh". Once we get to implementing this suggestion, let's stick with `300` for the barrier.

PR #245, opened 2018-10-23
Flag for low expiry cache time

rebeccahum:
Per #81, we should flag for expiry times less than 300 (5 minutes).

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Cache/LowExpiryCacheTimeSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Cache/LowExpiryCacheTimeSniff.php
@@ -1,0 +1,89 @@
+<?php
+/**
+ * WordPressVIPMinimum Coding Standard.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
+
+namespace WordPressVIPMinimum\Sniffs\Cache;
+
+use WordPress\AbstractFunctionParameterSniff;
+
+/**
+ * This sniff throws a warning when low cache times are set.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ *
+ * @since 0.4.0
+ */
+class LowExpiryCacheTimeSniff extends AbstractFunctionParameterSniff {
+
+	/**
+	 * The group name for this group of functions.
+	 *
+	 * @var string
+	 */
+	protected $group_name = 'cache_functions';
+
+	/**
+	 * Functions this sniff is looking for.
+	 *
+	 * @var array The only requirement for this array is that the top level
+	 *            array keys are the names of the functions you're looking for.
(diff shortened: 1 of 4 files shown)


WordPressVIPMinimum/Sniffs/Cache/LowExpiryCacheTimeSniff.php
```diff
@@ -0,0 +1,91 @@
...
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 * @return int|void Integer stack pointer to skip forward or void to continue
+	 *                  normal file processing.
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		if ( false === isset( $parameters[4] ) ) {
+			// If no cache expiry time, bail.
```
GaryJones: If there's no cache time, what does it default to?

WordPressVIPMinimum/Sniffs/Cache/LowExpiryCacheTimeSniff.php
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		$time = $parameters[4]['raw'];
+
+		if ( false === is_numeric( $time ) ) {
+			// If using time constants, we need to convert to a number.
+			$time = str_replace( array_keys( $this->wp_time_constants ), array_values( $this->wp_time_constants ), $time );
+
+			if ( preg_match( '#^[\s\d+*/-]+$#', $time ) > 0 ) {
+				$time = eval( "return $time;" ); // @codingStandardsIgnoreLine - No harm here.
```
GaryJones: Use `// phpcs:ignore Name.Of.Sniff.Check -- reasons` format, instad of // `@codingStandardsIgnoreLine`.

WordPressVIPMinimum/Tests/Cache/LowExpiryCacheTimeUnitTest.inc
```diff
@@ -0,0 +1,40 @@
+<?php
+
+$data = [ 'test', 'banana', 'ice cream', '123' ];
+$testing = 'test_key';
+$group = 'test_group';
+
```
GaryJones: Need a test for when no 4th parameter is given.

WordPressVIPMinimum/Sniffs/Cache/LowExpiryCacheTimeSniff.php
GaryJones: OK - add that example use case to the comment, so the next person doesn't question _why_ the code is like this.

PR #246, opened 2018-10-24
IncludingFile Sniff is more accurate

rebeccahum:
Including files with external URL or custom function (which may possibly return external URL) returned inaccurate warnings.  With this snippet as an example test:

```
require_once(wpcom_vip_theme_url('/my-theme/my-file.php', 'my-other-theme'));

require_once( 'http://example.com/my-file.php' );
```

Before:

```
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AND 1 WARNING AFFECTING 2 LINES
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
--------------------------------------------------------------------------------
 3 | WARNING | File inclusion using custom constant (`wpcom_vip_theme_url`).
   |         | Probably needs manual inspection.
   |         | (WordPressVIPMinimum.Files.IncludingFile.IncludingFile)
 5 | WARNING | Absolute include path must be used. Use
   |         | `get_template_directory()`, `get_stylesheet_directory()` or
   |         | `plugin_dir_path()`.
   |         | (WordPressVIPMinimum.Files.IncludingFile.IncludingFile)
--------------------------------------------------------------------------------

------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

After:

```
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
--------------------------------------------------------------------------------
 3 | WARNING | File inclusion using custom function ( `wpcom_vip_theme_url()`
   |         | ). Must return local file source, as external URLs are
   |         | prohibited on WordPress VIP. Probably needs manual
   |         | inspection.
   |         | (WordPressVIPMinimum.Files.IncludingFile.IncludingFile)
 5 | WARNING | Include path must be local file source, external URLs are
   |         | probibited on WordPress VIP.
   |         | (WordPressVIPMinimum.Files.IncludingFile.IncludingFile)
--------------------------------------------------------------------------------
```

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
@@ -123,7 +123,9 @@ class IncludingFileSniff implements Sniff {
 				return;
 			}
 
-			if ( 1 === preg_match( '/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $tokens[ $nextToken ]['content'] ) ) {
+			$nextNextToken = $phpcsFile->findNext( Tokens::$emptyTokens, ( $nextToken + 1 ), null, true, null, true );
+
+			if ( 1 === preg_match( '/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $tokens[ $nextToken ]['content'] ) && T_OPEN_PARENTHESIS !== $tokens[ $nextNextToken ]['code'] ) {
 				// The construct is using custom constant, which needs manual inspection.
 				$phpcsFile->addWarning( sprintf( 'File inclusion using custom constant (`%s`). Probably needs manual inspection.', $tokens[ $nextToken ]['content'] ), $nextToken, 'IncludingFile' );
 				return;
@@ -139,15 +141,19 @@ class IncludingFileSniff implements Sniff {
 				return;
 			}
 
-			$nextNextToken = $phpcsFile->findNext( Tokens::$emptyTokens, ( $nextToken + 1 ), null, true, null, true );
 			if ( T_OPEN_PARENTHESIS === $tokens[ $nextNextToken ]['code'] ) {
-				$phpcsFile->addWarning( sprintf( 'File inclusion using custom function ( `%s()` ). Probably needs manual inspection.', $tokens[ $nextToken ]['content'] ), $nextToken, 'IncludingFile' );
+				$phpcsFile->addWarning( sprintf( 'File inclusion using custom function ( `%s()` ). Must return local file source, as external URLs are prohibited on WordPress VIP. Probably needs manual inspection.', $tokens[ $nextToken ]['content'] ), $nextToken, 'IncludingFile' );
 				return;
 			}
 
 			$phpcsFile->addError( 'Absolute include path must be used. Use `get_template_directory()`, `get_stylesheet_directory()` or `plugin_dir_path()`.', $nextToken, 'IncludingFile' );
 			return;
 		} else {
+			if ( T_CONSTANT_ENCAPSED_STRING === $tokens[ $nextToken ]['code'] && filter_var( str_replace( [ '"', "'" ], '', $tokens[ $nextToken ]['content'] ), FILTER_VALIDATE_URL ) ) {
+				$phpcsFile->addError( 'Include path must be local file source, external URLs are probibited on WordPress VIP.', $nextToken, 'IncludingFile' );
+				return;
+			}
+
 			$phpcsFile->addError( 'Absolute include path must be used. Use `get_template_directory()`, `get_stylesheet_directory()` or `plugin_dir_path()`.', $nextToken, 'IncludingFile' );
 			return;
 		}
--- a/WordPressVIPMinimum/Tests/Files/IncludingFileUnitTest.inc
+++ b/WordPressVIPMinimum/Tests/Files/IncludingFileUnitTest.inc
@@ -11,3 +11,13 @@ include ( MY_CONSTANT . "my_file.php" ); // Warning.
 require_once ( MY_CONSTANT . "my_file.php" ); // Warning.
 
 include( locate_template('index-loop.php') ); // Ok.
+
+include( 'http://www.google.com/bad_file.php' ); // Bad.
+
+include("http://www.google.com/bad_file.php"); // Bad.
+
+require_once( custom_function( 'test_string.php' ) ); // Warning.
+
+function custom_function( $string ) {
+	return 'http://www.google.com/' . $string;
+}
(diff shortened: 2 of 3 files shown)


WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -139,15 +141,19 @@ public function process( File $phpcsFile, $stackPtr ) {
...
 
 			$phpcsFile->addError( 'Absolute include path must be used. Use `get_template_directory()`, `get_stylesheet_directory()` or `plugin_dir_path()`.', $nextToken, 'IncludingFile' );
 			return;
 		} else {
+			if ( T_CONSTANT_ENCAPSED_STRING === $tokens[ $nextToken ]['code'] && filter_var( str_replace( [ '"', "'" ], '', $tokens[ $nextToken ]['content'] ), FILTER_VALIDATE_URL ) ) {
+				$phpcsFile->addError( 'Include path must be local file source, external URLs are prohibited on WordPress VIP.', $nextToken, 'IncludingFile' );
+				return;
+			}
+
 			$phpcsFile->addError( 'Absolute include path must be used. Use `get_template_directory()`, `get_stylesheet_directory()` or `plugin_dir_path()`.', $nextToken, 'IncludingFile' );
```
GaryJones: Are there still test cases where this `addError()` is hit?

WordPressVIPMinimum/Tests/Files/IncludingFileUnitTest.inc
```diff
@@ -10,4 +10,14 @@ include ( MY_CONSTANT . "my_file.php" ); // Warning.
 
 require_once ( MY_CONSTANT . "my_file.php" ); // Warning.
 
-include( locate_template('index-loop.php') ); // Ok.
\ No newline at end of file
+include( locate_template('index-loop.php') ); // Ok.
+
+include( 'http://www.google.com/bad_file.php' ); // Bad.
+
+include("http://www.google.com/bad_file.php"); // Bad.
```
GaryJones: Maybe change this to `include_once`, as that seems to be missing.

WordPressVIPMinimum/Tests/Files/IncludingFileUnitTest.inc
```diff
@@ -10,4 +10,14 @@ include ( MY_CONSTANT . "my_file.php" ); // Warning.
 
 require_once ( MY_CONSTANT . "my_file.php" ); // Warning.
 
-include( locate_template('index-loop.php') ); // Ok.
\ No newline at end of file
+include( locate_template('index-loop.php') ); // Ok.
+
+include( 'http://www.google.com/bad_file.php' ); // Bad.
```
GaryJones: Can you update the comments to explain _why_ each of these test cases are OK/Bad/Warning?

WordPressVIPMinimum/Tests/Files/IncludingFileUnitTest.inc
```diff
@@ -10,4 +10,14 @@ include ( MY_CONSTANT . "my_file.php" ); // Warning.
```
GaryJones: How about:

```php
require dirname( __FILE__ ) . '/my_file.php' );
require dirname( __FILE__, 2 ) . '/my_file.php' );
require dirname( dirname( __FILE__ ) ) . '/my_file.php' );
require dirname( dirname( __DIR__ ) ) . '/my_file.php' );
require '../my_file.php' );
require '../../my_file.php' );
```

?

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -123,7 +123,9 @@ public function process( File $phpcsFile, $stackPtr ) {
 				return;
 			}
 
-			if ( 1 === preg_match( '/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $tokens[ $nextToken ]['content'] ) ) {
+			$nextNextToken = $phpcsFile->findNext( Tokens::$emptyTokens, ( $nextToken + 1 ), null, true, null, true );
+
+			if ( 1 === preg_match( '/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $tokens[ $nextToken ]['content'] ) && T_OPEN_PARENTHESIS !== $tokens[ $nextNextToken ]['code'] ) {
```
GaryJones: Which test case goes with this check?

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -123,7 +123,9 @@ public function process( File $phpcsFile, $stackPtr ) {
 				return;
 			}
 
-			if ( 1 === preg_match( '/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $tokens[ $nextToken ]['content'] ) ) {
+			$nextNextToken = $phpcsFile->findNext( Tokens::$emptyTokens, ( $nextToken + 1 ), null, true, null, true );
```
GaryJones: Instead of searching for the next next token, could we see if the current next token is a function call (type is `T_FUNCTION` or something?), and skip over it that way?

rebeccahum: Next best thing to do would be to add `T_COMMENT` as part of what not to look for I think.

GaryJones: Maybe abstract from WPCS's AbstractFunctionRestrictions, and then use the method [here](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/AbstractFunctionRestrictionsSniff.php#L205) ?

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
rebeccahum: `include ( MY_CONSTANT . "my_file.php" ); // Custom constant.`

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
rebeccahum: I can do that but that would still end up returning `true` for constants it looks like, unless I'm _doing_it_wrong.

rebeccahum: I can do that but that would still end up returning `true` for constants it looks like, so we'd still need to search for the nextNextToken.

WordPressVIPMinimum/Tests/Files/IncludingFileUnitTest.inc
```diff
@@ -1,13 +1,27 @@
 <?php
```
GaryJones: _Much_ nicer arrangement of test cases 👍 - far easier to read.

Issue #156, opened 2018-02-20
Catch usage of the_sub_field()


gudmdharalds:
`the_sub_field()` (part of Advanced Custom Fields) does not escape output -- we should catch usage of this function and warn that the output will go out unescaped.

See: https://github.com/elliotcondon/acf/blob/master/core/api.php#L681

gudmdharalds:
Same applies to `the_field()`: https://github.com/elliotcondon/acf/blob/master/core/api.php#L329

rebeccahum:
Closing per #248

PR #248, opened 2018-10-24
Add ACF templating function to warnings

rebeccahum:
Per #156, ACF's `the_field()` and `the_sub_field()` do not escape outputted.

### DIFF
--- a/WordPressVIPMinimum/Tests/VIP/RestrictedFunctionsUnitTest.inc
+++ b/WordPressVIPMinimum/Tests/VIP/RestrictedFunctionsUnitTest.inc
@@ -42,6 +42,8 @@ is_multi_author(); // Bad. Warning.
 
 opcache_reset(); // Bad.
 
-// Warning - ACF.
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Tests/VIP/RestrictedFunctionsUnitTest.inc
```diff
@@ -40,4 +40,8 @@ dbDelta(); // Bad. Warning.
 
 is_multi_author(); // Bad. Warning.
 
-opcache_reset(); // Bad.
\ No newline at end of file
+opcache_reset(); // Bad.
+
+// Warning - ACF.
+the_field( 'field' );
+the_sub_field( 'field' );
```
GaryJones: Add test cases showing how it should be done i.e. echo and escaping.

Also, keep the newline at the end of the file, per WPCS (and avoid the red marker here). Your editor can be set to follow that, though an `.editorconfig` should be added as well.

PR #257, opened 2018-10-31
Add StripTags sniff and amend rulesets and existing sniffs

### LINKED ISSUE #133: Suggest using `wp_kses` over `strip_tags` in case `allowable_tags` are being used

rebeccahum:
This resolves #133.

What this PR does:
* add a StripTags sniff
* remove the WordPress.AlternativeFunctions rules for strip tags
* remove strip_tags from RestrictedFunctions sniff

### DIFF
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -147,10 +147,9 @@
 		<type>warning</type>
 		<severity>7</severity>
 	</rule>
-	<rule ref="WordPressVIPMinimum.VIP.RestrictedFunctions.strip_tags_strip_tags">
+	<rule ref="WordPressVIPMinimum.Functions.StripTags">
 		<type>warning</type>
 		<severity>5</severity>
-		<message>We recommend using wp_kses() or wp_kses_post() instead https://codex.wordpress.org/Function_Reference/wp_kses</message>
 	</rule>
 	<rule ref="WordPress.WP.PostsPerPage.posts_per_page_posts_per_page">
 		<type>warning</type>
--- a/WordPressVIPMinimum/Sniffs/Functions/StripTagsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/StripTagsSniff.php
@@ -50,15 +50,15 @@ class StripTagsSniff extends AbstractFunctionParameterSniff {
 	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
 		if ( 1 === count( $parameters ) ) {
 			$this->phpcsFile->addWarning(
-					sprintf( '`strip_tags()` does not strip CSS and JS in between the script and style tags. Use `wp_strip_all_tags()` to strip all tags.', $matched_content ),
-					$stackPtr,
-					'StripTagsOneParameter'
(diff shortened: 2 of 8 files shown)


GaryJones:
Background: [WPCS code](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/46d42828ce7355d8b3776e3171f2bda892d179b4/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php#L96-L103) added a straight replacement suggestion if `strip_tags()` was used. `wp_strip_all_tags()` is recommended, as it properly strips all HTML tags, including _the contents_ of `script` and `style` elements.

#133 says that using `strip_tags( $some_variable, '<p><a><iframe><strong>' );` would be wrong. @david-binda is this because `wp_kses()` does a better job, security wise, than `strip_tags()`?

The first warning in this PR basically matches that from WPCS.

The second recomends `wp_kses()` when the second arg is used in the `strip_tags()` call.

cc @jrfnl in case this would be suitable for WPCS.

GaryJones:
@jrfnl Thanks :-)

> Adding a suggestion for using `wp_kses()` in those cases could be an addition to the existing WPCS sniff. No new sniff needed for that.

@rebeccahum Based on the above, I think it would be better see about opening a PR on WPCS for the same end result, and if successful. closing out this PR unmerged.

jrfnl:
> Based on the above, I think it would be better see about opening a PR on WPCS for the same end result

I'm not sure such a PR should be accepted to WPCS as `wp_kses()` is quite a different function from `strip_tags()`. `wp_kses()` is only intended for output escaping, while `(wp_)strip_(all_)tags()` can also be used as part of input sanitization.

david-binda:
>  @david-binda is this because wp_kses() does a better job, security wise, than strip_tags()?

Well, the issue is two fold. First of all, `wp_strip_all_tags` does strip *all* tags and there is no way to preserve any tags. The other difference in between `strip_tags` and `wp_strip_all_tags` is that the later removes content enclosed by `<script>` and `<style>` params:

```
wp> $text = '<p><strong><a href="#somelink">Link</a></strong><iframe src="https://example.com"></iframe><img src="./images/image.gif"/></p><script>alert("haxxored");</script>';
wp> strip_tags( $text, '<p><a><iframe><strong><img>' );
string(144) "<p><strong><a href="#somelink">Link</a></strong><iframe src="https://example.com"></iframe><img src="./images/image.gif"/></p>alert("haxxored");"
wp> wp_strip_all_tags( $text );
string(4) "Link"
```

More obvious example:

```
wp> $text = '<script>alert("haxxored")</script>';
string(34) "<script>alert("haxxored")</script>"
wp> strip_tags( $text );
string(17) "alert("haxxored")"
wp> wp_strip_all_tags( $text );
string(0) ""
```

So, in case we want to preserve some HTML in WordPress.com VIP world, we would use `wp_kses` which allows us to define a set of HMTL elements, including attributes, and http protocols.

PR #269, opened 2018-11-15
Add check for duplicate before throwing warning in VariableAnalysis Sniff

### LINKED ISSUE #262: Duplicate "unused variable" warning
## Bug Description
An "unused variable" warning appears twice.

## Minimal Code Snippet
```php
<?php
	function test() {
		try {
			echo 'hi';
		} catch ( Exception $e ) {
			echo 'done';
		}
	}

```

## Error Code
```
$ vendor/bin/phpcs ~/test.php --sniffs=WordPressVIPMinimum.Variables.VariableAnalysis

FILE: Users/schreipa/test.php
-----------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 1 LINE
-----------------------------------------------------------------------------------------------------
 5 | WARNING | Unused variable `$e`.
   |         | (WordPressVIPMinimum.Variables.VariableAnalysis.UnusedVariable)
 5 | WARNING | Unused variable `$e`.
   |         | (WordPressVIPMinimum.Variables.VariableAnalysis.UnusedVariable)
-----------------------------------------------------------------------------------------------------
```

## Environment

| Question               | Answer
| PHP version             | 7.2.2
| PHP_CodeSniffer version | 3.3.2
| WPCS version            | 0.3.0
| WPCS install type       | composer project local

rebeccahum:
Fixes #262.


GaryJones:
Applies generally, but I'd like to see the commit message:

 - include a description of the problem that this fix addresses.
 - include some background on why this fix works - how is it that these variables can be compared to avoid the duplicate.
 - include a reference to the ticket that it addresses (`See #262` or `Fixes #262`)

The commit should be self-contained in terms of the meta info that is provided in the message - if I look at the blame in my IDE, or command line, or Source Tree / GitHub Desktop, etc., I want to be able to see sufficient context to understand and appreciate the change. Let's not just presume that GitHub is the only source of info here; if we moved to another Git provider, the commit message should contain all the reasoning.

rebeccahum:
I think I'll leave that for #270 if that's alright with you.

GaryJones:
In which case, maybe this should be considered a duplicate of #270?

If #270 hides the reporting of the `$e`, then presumably it wouldn't report it twice, which is what this PR fixes. Is that right?

If so, let's get #270 in, and check if #262 still occurs.

There may well be other instances of duplicate variable violations, but we can address them when needed.

Issue #211, opened 2018-08-28
Raise minimum version of PHP?


GaryJones:
In [dropping support for PHPCS 2.x](https://github.com/Automattic/VIP-Coding-Standards/issues/190), the minimum version of PHP was raised to [PHP >=5.4](https://github.com/Automattic/VIP-Coding-Standards/blob/2672675106afb5b85f9d338aa2c7a57f8dc0222d/composer.json#L13).

@tomjn has proposed raising this requirement for running VIPCS to at least "a supported version PHP version like 5.6".

Thoughts?

tomjn:
imo PHPCS should run on PHP 7.2, even if the code targets unsupported versions of PHP. PHPCompat and PHPUnit can test the plugins and themes in 5.2/5.4 if it's necessary, but that doesn't require PHPCS to run on those versions too

If necessary, they can add the PHPCompatibility ruleset as a dependency to add another layer of protection in addition to VIPCS

GaryJones:
For _VIPCS_ Travis, we run PHPCS against the VIPCS code only at 7.2:

 - [current Travis config](https://github.com/Automattic/VIP-Coding-Standards/blob/2672675106afb5b85f9d338aa2c7a57f8dc0222d/.travis.yml#L35)
 - [open PR to improve Travis](https://github.com/Automattic/VIP-Coding-Standards/pull/210/files#diff-354f30a63fb0907d4ad57269548329e3R88)

For VIPCS unit tests in Travis, it needs a [`composer install`](https://github.com/Automattic/VIP-Coding-Standards/blob/2672675106afb5b85f9d338aa2c7a57f8dc0222d/.travis.yml#L69), though there _may_ be a way for it to use any of the desired versions PHP, while ignoring any PHP requirement set for VIPCS itself. Maybe [`platform`](https://getcomposer.org/doc/06-config.md#platform) (which WP-CLI [uses](https://github.com/wp-cli/wp-cli/blob/8d59b90b9a6c9f2667923971a0292bf8e020036d/composer.json#L29))?

rinatkhaziev:
PHP5.6 will be officially EOL'd soon (Dec 3, 2018): http://php.net/supported-versions.php

Personally, I think the requirement should be at least 7.0, preferably 7.2 since it's what we run.

justnorris:
While I agree that every web server and dev should have PHP7+ on their machine, I think that we should require a version not for the name of the version, but for the features we rely upon.

[...]

PR #270, opened 2018-11-15
Hide catching exceptions

### LINKED ISSUE #114: Unused variable $e.
We really should not report on `$e` in try...catch

```
} catch ( Exception $e ) {
```

rebeccahum:
Fixes #114. This is a reboot of #144.

### DIFF
--- a/WordPressVIPMinimum/Tests/Variables/VariableAnalysisUnitTest.php
+++ b/WordPressVIPMinimum/Tests/Variables/VariableAnalysisUnitTest.php
@@ -31,9 +31,16 @@ class VariableAnalysisUnitTest extends AbstractSniffUnitTest {
 	 * @return array <int line number> => <int number of warnings>
 	 */
 	public function getWarningList() {
-		return array(
+		$warningList = [
 			5 => 2,
-		);
+		];
+	
+ 		// PHP prior to version 7.x does not properly process the $e.
+		if ( true === version_compare( PHP_VERSION, '7.0.0', '>' ) ) {
+			$warningList[18] = 2;
+		}
+
+ 		return $warningList;
 	}
 
 }
--- a/WordPressVIPMinimum/Tests/Variables/VariableAnalysisUnitTest.php
+++ b/WordPressVIPMinimum/Tests/Variables/VariableAnalysisUnitTest.php
@@ -31,16 +31,10 @@ class VariableAnalysisUnitTest extends AbstractSniffUnitTest {
 	 * @return array <int line number> => <int number of warnings>
 	 */
 	public function getWarningList() {
-		$warningList = [
+		return [
 			5 => 2,
+			18 => 2,
 		];
-
-		// PHP prior to version 7.x does not properly process the $e.
-		if ( true === version_compare( PHP_VERSION, '7.0.0', '>=' ) ) {
-			$warningList[18] = 2;
-		}
-
-		return $warningList;
 	}
 
 }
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Tests/Variables/VariableAnalysisUnitTest.php
```diff
@@ -31,9 +31,16 @@ public function getErrorList() {
 	 * @return array <int line number> => <int number of warnings>
 	 */
 	public function getWarningList() {
-		return array(
+		$warningList = [
 			5 => 2,
-		);
+		];
+	
+ 		// PHP prior to version 7.x does not properly process the $e.
```
GaryJones: There's a leading space at the start of this line.

WordPressVIPMinimum/Tests/Variables/VariableAnalysisUnitTest.php
```diff
@@ -31,9 +31,16 @@ public function getErrorList() {
...
+		if ( true === version_compare( PHP_VERSION, '7.0.0', '>' ) ) {
+			$warningList[18] = 2;
+		}
+
+ 		return $warningList;
```
GaryJones: There's a leading space at the start of this line.

WordPressVIPMinimum/Tests/Variables/VariableAnalysisUnitTest.php
```diff
@@ -31,9 +31,16 @@ public function getErrorList() {
...
 	 */
 	public function getWarningList() {
-		return array(
+		$warningList = [
 			5 => 2,
-		);
+		];
+	
+ 		// PHP prior to version 7.x does not properly process the $e.
+		if ( true === version_compare( PHP_VERSION, '7.0.0', '>' ) ) {
```
GaryJones: Should this be `'>='`, so that it matches exactly `7.0.0` as well?

WordPressVIPMinimum/Tests/Variables/VariableAnalysisUnitTest.php
GaryJones: Is there any reference that could be added, to find out more about why it doesn't properly process the `$e`?

Could the code that does the processing be updated, so it does?

rebeccahum: With VIP running a later version, would it even be applicable to check that PHP is < 7?

GaryJones: This repo supports [down to PHP 5.6](https://github.com/Automattic/VIP-Coding-Standards/blob/master/composer.json#L18) (related: https://github.com/Automattic/VIP-Coding-Standards/issues/211).

Until the level is raised to 7.0+ (or later), then yes, it should be checked.

The version that these sniffs are run at, and the version that the code is expected to run at, can be two different versions.

rebeccahum: > Is there any reference that could be added, to find out more about why it doesn't properly process the $e?

I couldn't find any reference on _why_ and speaking to @david-binda, he doesn't recall why either.

rebeccahum: Update: I don't believe this is an issue anymore since it appears that the version check was added because it was failing the tests.

PR #281, opened 2018-11-28
Flag filters for changing timeouts

### LINKED ISSUE #45: Flag code increasing HTTP request timeouts
> We strongly discourage using a timeout greater than 3 seconds since remote requests require that the user wait for them to complete before the rest of the page will load.

Per https://vip.wordpress.com/documentation/vip/best-practices/fetching-remote-data/

We have a lot of helper functions available for developers, and we are trying to actively report on https://opengrok.a8c.com/source/xref/vip/plugins/vip-helper.php#172 any timeout higher that 3s and in some instances we are blocking timeout higher than 10s ( https://opengrok.a8c.com/source/xref/vip/plugins/vip-helper.php#169 ) or even 3s ( https://opengrok.a8c.com/source/xref/vip/plugins/vip-helper.php#617 )

But there are other means for bypassing the timeout limits. Eg.: `http_request_timeout` or `http_request_args ` filters.

We might want to flag those for manual inspection during the code review.

rebeccahum:
Fixes #45.  

Also changes RestrictedHookSniff to have different error types on a hook basis.  I chose a severity of 5 since it matches the other performance issues flagged with a warning of 5.

### DIFF
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -159,10 +159,6 @@
 		<severity>5</severity>
 		<message>Having more than 100 posts returned per page can lead to severe performance problems.</message>
 	</rule>
-	<rule ref="WordPressVIPMinimum.Filters.RestrictedHook.HighTimeout">
-		<type>warning</type>
-		<severity>5</severity>
-	</rule>
 	<rule ref="WordPressVIPMinimum.Classes.RestrictedExtendClasses.wp_cli">
 		<type>warning</type>
 		<severity>5</severity>
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -159,6 +159,10 @@
 		<severity>5</severity>
 		<message>Having more than 100 posts returned per page can lead to severe performance problems.</message>
 	</rule>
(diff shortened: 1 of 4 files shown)


WordPress-VIP-Go/ruleset.xml
```diff
@@ -159,6 +159,10 @@
 		<severity>5</severity>
 		<message>Having more than 100 posts returned per page can lead to severe performance problems.</message>
 	</rule>
+	<rule ref="WordPressVIPMinimum.Filters.RestrictedHook.HighTimeout">
```
GaryJones: These four lines do nothing - the type is already set as a `Warning`, and the severity of 5 is the default anyway.

rebeccahum: @GaryJones Then should I just add the ruleset `WordPressVIPMinimum.Filters.RestrictedHook` as a whole or is it not necessary at all?

GaryJones: It shouldn't be necessary, as [line 6](https://github.com/Automattic/VIP-Coding-Standards/pull/281/files#diff-0668f43b46ab1d5d149e12c1529486b8R6) adds the whole of `WordPressVIPMinimum` ruleset / sniffs anyway.

IMO the only thing that should appear here, are _amendments_ - a change of message or severity or type, for VIP Go, compared to the attributes set in the Sniff which apply to WPCOM.

PR #283, opened 2018-11-28
Fix error message for FetchRemoteDataSniff

### LINKED ISSUE #243: WordPressVIPMinimum.VIP.FetchingRemoteData.fileGetContentsUknown buggy error message
## Bug Description
Seeing a buggy error message here:

> %s() is uncached. If this is being used to query a remote file please use wpcom_vip_file_get_contents() instead. If it\'s used for a local file please use WP_Filesystem instead. Read more here: **https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/** 

stemming from the sniff `WordPressVIPMinimum.VIP.FetchingRemoteData.fileGetContentsUknown` when it detects `file_get_contents()`

rebeccahum:
Fixes #243.

Also uses `addWarning()` without `sprintf()` since that's improper.


WordPressVIPMinimum/Sniffs/Variables/VariableAnalysisSniff.php
```diff
@@ -728,8 +728,7 @@ protected function checkForFunctionPrototype(
 				// We haven't been defined by this point.
 				$phpcsFile->addWarning( "Variable `%s` is undefined.", $stackPtr,
 					'UndefinedVariable',
-					array( "\${$varName}" ),
-					'WordPressVIPMinimum.Variables.VariableAnalysis'
+					array( "\${$varName}" )
```
GaryJones: These changes were already merged via a different PR. Please pull `master` into this PR's branch, so that we don't re-commit the changes and possibly lose the commit message and reference from the first effort.

WordPress-VIP-Go/ruleset.xml
```diff
@@ -50,7 +50,6 @@
 	<rule ref="WordPressVIPMinimum.VIP.FetchingRemoteData.fileGetContentsUknown">
 		<type>error</type>
 		<Severity>5</Severity>
-		<message>%s() is uncached. If this is being used to query a remote file please use wpcom_vip_file_get_contents() instead. If it's used for a local file please use WP_Filesystem instead. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
```
GaryJones: This contains a documentation link that isn't present in the message in the sniff. Is it still of value to keep the link, either in the sniff (for WPCOM + Go), or this ruleset (just Go)?

rebeccahum: I don't think it's of value to keep the link since it's marked as "VIP Go Platform Specific".  If anything, I'm thinking that we should use https://codex.wordpress.org/Filesystem_API.

PR #286, opened 2018-11-28
Remove FlushRewriteRuleSniff since it is already being checked in another sniff.

### LINKED ISSUE #279: Duplicate FlushRewriteRules checks
- [`WordPressVIPMinimum.VIP.RestrictedFunctions.rewrite_rules_flush_rewrite_rules`](https://github.com/Automattic/VIP-Coding-Standards/blob/ebfb056875e0ff615d6d7cb8e9562c511d6c9588/WordPressVIPMinimum/Sniffs/VIP/RestrictedFunctionsSniff.php#L66-L72)
 - [`FlushRewriteRulesSniff`](https://github.com/Automattic/VIP-Coding-Standards/blob/master/WordPressVIPMinimum/Sniffs/VIP/FlushRewriteRulesSniff.php)

Both of these seem to target the same violation, which may result in duplicate messages.

rebeccahum:
Fixes #279.


GaryJones:
This is a breaking change. We should look at whether it is worth doing for 0.4.0, or wait until 0.5.0.

GaryJones:
@rebeccahum The integration test contains:

```php
$wp_rewrite->flush_rules(); // Bad.
```

With the `WordPressVIPMinimum.VIP.FlushRewriteRules` present, this call is caught and reports one violation.

With this PR, it is not - it seems that the `FlushRewriteRulesSniff` is not equivalent to the `WordPressVIPMinimum.VIP.RestrictedFunctions.rewrite_rules_flush_rewrite_rules` check.

So, either the `WordPressVIPMinimum.VIP.RestrictedFunctions` also needs to look for this `$wp_rewrite->flush_rules()` call, or the bit of that checks for `flush_rewrite_rules` should be removed and the FlushRewriteRulesSniff be left in place.

rebeccahum:
> either the WordPressVIPMinimum.VIP.RestrictedFunctions also needs to look for this $wp_rewrite->flush_rules() call, 

I chose this approach so if we need to add other restricted functions belonging to certain classes then going forward, then we can do so.

WordPressVIPMinimum/Sniffs/VIP/RestrictedFunctionsSniff.php
```diff
@@ -244,4 +255,49 @@ public function getGroups() {
 
 		return $groups;
 	}
+
+	/**
+	 * Verify is the current token is a function call.
```
GaryJones: Seems to be an extra `is` in the description.

WordPressVIPMinimum/Sniffs/VIP/RestrictedFunctionsSniff.php
```diff
@@ -244,4 +255,49 @@ public function getGroups() {
...
+	 *
+	 * @param int $stackPtr The position of the current token in the stack.
+	 *
+	 * @return bool
+	 */
+	public function is_targetted_token( $stackPtr ) {
```
GaryJones: This method doesn't seem to be in use?

rebeccahum: @GaryJones See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/00f5d8989f4c6b6c592c54e202aa7ef608988b88/WordPress/AbstractFunctionRestrictionsSniff.php#L213

GaryJones: 1. Does this method differ from the method in the parent class?

  If not, adding it here as well adds nothing.

  If it does differ, then it should be clearly identified how it differs.

2. Where is this method called in this class?

rebeccahum: > 1. Does this method differ from the method in the parent class?

Yes -- it differs by checking if the restricted function in question is part of a specific class before flagging it.

> 2. Where is this method called in this class?

Right here in the parent class: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/00f5d8989f4c6b6c592c54e202aa7ef608988b88/WordPress/AbstractFunctionRestrictionsSniff.php#L199

rebeccahum: > 1. Does this method differ from the method in the parent class?

Yes -- it differs by checking if the restricted function in question is part of a specific class before flagging it.

> 2. Where is this method called in this class?

Right here in the parent class: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/00f5d8989f4c6b6c592c54e202aa7ef608988b88/WordPress/AbstractFunctionRestrictionsSniff.php#L199

GaryJones: Ah,got it, thanks.

GaryJones:
Audit updated.

Change log notes:

### Added
- `WordPressVIPMinimum.VIP.RestrictedFunctions.flush_rules_flush_rules` as part of removing `FlushRewriteRulesSniff`.


### Changed
- BREAKING: `WordPressVIPMinimum.VIP.RestrictedFunctions.rewrite_rules_flush_rewrite_rules` to  `WordPressVIPMinimum.VIP.RestrictedFunctions.flush_rewrite_rules_flush_rewrite_rules` as part of removing `FlushRewriteRulesSniff`.

### Removed
 - BREAKING: `WordPressVIPMinimum.VIP.FlushRewriteRules` sniff, since it overlapped with checks in ` WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff`.

PR #292, opened 2018-12-04
Add PHP Syntax Error sniff to WordPressVIPMinimum

### LINKED ISSUE #289: Syntax Error Sniff

The `WordPress-Extra` sniffs include syntax error detection, but since we don't extend that (just cherry pick out various WP sniffs via `ruleset.xml`, we are missing syntax errors.

test.php
```
<?php

thisisasyntaxerror
```

Running the ruleset produces no errors though:

```
$ phpcs --standard=WordPressVIPMinimum test.php
$
```

Syntax errors are caught by the `WordPressVIPMinimum` and `WordPress-VIP-Go` rulesets.

Any PHP syntax error.

Valid PHP code.

## Additional context

We're missing this:

```
<rule ref="Generic.PHP.Syntax"/>
```

Right now we're picking certain rules out of the WordPress rulesets - would it make more sense to include the full ruleset and then exclude rules we don't want?

nickdaugherty:
Has been inadvertently omitted until now, but it's arguably the most important sniff to have.

Fixes #289


WordPressVIPMinimum/ruleset.xml
```diff
@@ -8,6 +8,8 @@
 
 	<autoload>./PHPCSCompatibility.php</autoload>
 
+	<rule ref="Generic.PHP.Syntax" />
```
GaryJones: There should be no space before the `/>`. This is XML, not XHTML.

GaryJones:
Nothing that this change will affect both `WordPressVIPMinimum` and `WordPress-VIP-Go` rulesets (since the latter inherits the former).

gudmdharalds:
> @gudmdharalds Since the vip-ci-bot already does php -l, should anything change there once this commit is used by the bot, to avoid duplicate reporting?

Yep, we will have to skip this sniff. There is also another sniff to ignore, so I will add an option to `vip-go-ci` to make sure we can skip custom sniffs.

gudmdharalds:
> @gudmdharalds Since the vip-ci-bot already does php -l, should anything change there once this commit is used by the bot, to avoid duplicate reporting?

Yep, we will have to skip this sniff. There is also another sniff to ignore, so I will add an option to `vip-go-ci` to make sure we can skip custom sniffs.

gudmdharalds:
Yep, we will have to skip this sniff. There is also another sniff to ignore, so I will add an option to `vip-go-ci` to make sure we can skip custom sniffs.

gudmdharalds:
> @gudmdharalds Since the vip-ci-bot already does php -l, should anything change there once this commit is used by the bot, to avoid duplicate reporting?

Yep, we will have to skip this sniff. There is also another sniff to ignore, so I will add an option to `vip-go-ci` to make sure we can skip custom sniffs.

gudmdharalds:
> @gudmdharalds Since the vip-ci-bot already does php -l, should anything change there once this commit is used by the bot, to avoid duplicate reporting?

Yep, we will have to skip this sniff. There is also another sniff to ignore, so I will add an option to `vip-go-ci` to make sure we can skip custom sniffs.

gudmdharalds:
This is the feature: https://github.com/Automattic/vip-go-ci/pull/31 -- aiming to deploy this after the weekend.

nickdaugherty:
@gudmdharalds is Automattic/vip-go-ci#31 a blocker to merging this?

gudmdharalds:
> @gudmdharalds is Automattic/vip-go-ci#31 a blocker to merging this?

Nope. Feel free to merge it.

PR #310, opened 2018-12-19
AlwaysReturnSniff will trigger errors unless callback args is pass by reference

### LINKED ISSUE #293: Consider making AlwaysReturnSniff raise errors instead of warnings

Incorrect returns from filters can ruin your day :)

Currently `return` problems from WP filters are flagged as Warnings, but I suggest they be raised to Error given their potential impacts and the unlikelihood of _actually_ needing to not return from inside a WP filter.

I guess anywhere that's passing by reference would need to be omitted?

Any WP filter that doesn't explicitly return a value.

Filters that _always_ return a value, and (I guess) filters that rely on passing by reference.

rebeccahum:
Fixes #293.

* If no return, it will trigger an error instead of a warning.
* If a callback arg is being passed in by reference, it should not trigger.

### DIFF
--- a/WordPressVIPMinimum/Tests/Filters/AlwaysReturnUnitTest.inc
+++ b/WordPressVIPMinimum/Tests/Filters/AlwaysReturnUnitTest.inc
@@ -1,6 +1,6 @@
 <?php
 
-function good_example_function() {
+function good_example_function() { // Ok.
 	if ( 1 === 0  ) {
 		if ( 1 === 1 ) {
 			return 'ahoj';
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Tests/Filters/AlwaysReturnUnitTest.inc
```diff
@@ -77,32 +77,68 @@ if ( ! class_exists( 'another_good_example_class' ) ) {
...
+			return 'ahoj';
+		} else {
+			return $test . 'hello';
+		}
+	}
+	// Missing universal return.
+}
+add_filter( 'bad_example_filter', 'bad_example_arg' );
+
+add_filter( 'bad_anonymous_example_arg', function( $input ) { // Error.
```
GaryJones: How is this different from `bad_example_closure_with_no_returns` on line 105?

rebeccahum: Whoops!  I'm not sure what I was thinking here...removing it now.  Nice catch!

PR #328, opened 2019-01-04
Combine SessionFunctionsSniff into RestrictedFunctionsSniff and SessionVariableSniff into RestrictedVariablesSniff

rebeccahum:
I don't think there is a need for a specific sniff for the sessions functions and variables when we already have the `RestrictedFunctionsSniff` and `RestrictedVariablesSniff`.


WordPressVIPMinimum/Tests/VIP/RestrictedFunctionsUnitTest.inc
```diff
@@ -137,3 +137,35 @@ get_category_by_slug(); // Ok - non-deprecated version of wpcom_vip_get_category
...
+session_save_path(); // Error.
+session_set_cookie_params(); // Error.
+session_set_save_handler(); // Error.
+session_start(); // Error.
+session_status(); // Error.
+session_unregister(); // Error.
+session_unset(); // Error.
+session_write_close(); // Error.
+
+// phpcs:set WordPressVIPMinimum.VIP.SessionFunctionsUsage exclude session
```
GaryJones: This L168-171 doesn't make sense now the `session_*()` functions have been moved to RestrictedFunctions?

WordPressVIPMinimum/Sniffs/VIP/SessionFunctionsUsageSniff.php
```diff
@@ -1,75 +0,0 @@
...
-	 *      'message'   => 'Use anonymous functions instead please!',
-	 *      'functions' => array( 'file_get_contents', 'create_function' ),
-	 *  )
-	 * )
-	 *
-	 * @return array
-	 */
-	public function getGroups() {
-		return [
-			'session' => [
```
GaryJones: Where has this array been moved to? I don't see any Restricted*Sniff updated in this PR.

screenshot 2019-01-04 at 23 55 57

This is why the [tests are failing](https://travis-ci.org/Automattic/VIP-Coding-Standards/jobs/475526016) :-)

GaryJones:
> I don't think there is a need for a specific sniff for the sessions functions and variables when we already have the RestrictedFunctionsSniff and RestrictedVariablesSniff.

Agreed.

Audit will need updating to adjust the path to the violations. I can do this.

Won't need to mark this as a breaking change, as the Session sniffs were only added in the current release cycle.

WordPressVIPMinimum/Sniffs/VIP/SessionFunctionsUsageSniff.php
rebeccahum: Sorry about that, this PR wasn't complete, I was mucking around with the rebasing at the time...

WordPressVIPMinimum/Sniffs/VIP/RestrictedVariablesSniff.php
```diff
@@ -46,6 +46,13 @@ public function getGroups() {
 					'$wpdb->usermeta',
 				],
 			],
+			'session' => [
+				'type'      => 'error',
+				'message'   => 'Due to using Batcache, server side based client related logic will not work, use JS instead.',
```
GaryJones: Was this message a copy and paste from other messages, or is it equally valid for this?

The previous message was simply `'Usage of $_SESSION variable is prohibited.'`

GaryJones:
Audit has been updated, though may need amending depending on the result of [this query](https://github.com/Automattic/VIP-Coding-Standards/pull/328#discussion_r245459543)

PR #341, opened 2019-01-08
Move FileSystemWritesDisallowSniff into RestrictedFunctionsSniff

rebeccahum:
It would be easier to keep track of everything since we moved the session sniffs into it too.


WordPressVIPMinimum/Sniffs/VIP/RestrictedFunctionsSniff.php
```diff
@@ -221,6 +221,46 @@ public function getGroups() {
 					'session_write_close',
 				],
 			],
+			'file_ops' => [
+				'type'      => 'error',
+				'message'   => 'Filesystem writes are forbidden, you should not be using %s()',
```
GaryJones: Please ensure messages end with full stops.

GaryJones:
This will need the audit to be updated.

Issue #334, opened 2019-01-06
Check if url_to_post_id() ever existed


GaryJones:
VIPCS contains references to a `url_to_post_id()` function. There is a very similarly named [`url_to_postid()`](https://developer.wordpress.org/reference/functions/url_to_postid/) function in WP core, but the former doesn't seem to exist outside of VIPCS and https://vip.wordpress.com/documentation/vip-go/uncached-functions/

@sboisvert Where did you find `url_to_post_id()`?

If it's not valid, then it can be removed.

GaryJones:
@sboisvert thinks he dreamt this up, so references can be removed.

PR #343, opened 2019-01-11
Remove non-existing function: url_to_post_id

WPprodigy:
Closes https://github.com/Automattic/VIP-Coding-Standards/issues/334

The rule name seems a bit convoluted, but I'm not sure if changing that is a good idea: https://github.com/Automattic/VIP-Coding-Standards/blob/b46ec66b8899d6ca8f83e9b0e70062ed6409c184/WordPress-VIP-Go/ruleset.xml#L128


GaryJones:
Audit has been updated.

PR #348, opened 2019-01-15
Remove fwrite and file_put_contents from RestrictedFunctions

rebeccahum:
In the `RestrictedFunctionsSniff`, we had `fwrite` and `file_put_contents` which was duplicating the effort of WPCS and producing duplicate error messages:

https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/ab999048a69e8614103af3fad4ccf4f124b9f520/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php#L91-L92

This PR:

* removes it from the `RestrictedFunctionsSniff`
* removed the duplicate VIPCS reference in the `WordPress-VIP-Go` ruleset for `fwrite`
* changes the reference to use WPCS for `file_put_contents` in the `WordPress-VIP-Go` ruleset
* adds both WPCS references to `WordPressVIPMinimum` ruleset for `fwrite` and `file_put_contents`
* changes the error message for filesystem write sniffs.

### DIFF
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -10,11 +10,63 @@
 
 
 	<!-- Things that probably won't work and needs a dev to review -->
-	<rule ref="WordPress.WP.AlternativeFunctions.file_system_read_fwrite">
-		<type>error</type>
+	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_delete">
+		<severity>6</severity>
+		<message>File system writes only work in /tmp/ and inside the /uploads/ folder on VIP Go. To do filesystem writes you must use the WP_Filesystem class, using functions such as %s() won't work or will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
+	</rule>
+	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_file_put_contents">
+		<severity>6</severity>
+		<message>File system writes only work in /tmp/ and inside the /uploads/ folder on VIP Go. To do filesystem writes you must use the WP_Filesystem class, using functions such as %s() won't work or will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
+	</rule>
+	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_flock">
+		<severity>6</severity>
+		<message>File system writes only work in /tmp/ and inside the /uploads/ folder on VIP Go. To do filesystem writes you must use the WP_Filesystem class, using functions such as %s() won't work or will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
+	</rule>
+	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fputcsv">
+		<severity>6</severity>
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
```diff
@@ -222,14 +222,12 @@ public function getGroups() {
 			],
 			'file_ops' => [
 				'type'      => 'error',
-				'message'   => 'Filesystem writes are forbidden, you should not be using %s().',
+				'message'   => 'Filesystem writes are forbidden, please do not use %s().',
 				'functions' => [
 					'delete',
-					'file_put_contents',
```
GaryJones: I think that at least a comment needs to go into these two removals, explaining why this would appear to be an incomplete list, compared to a docs page.

WordPress-VIP-Go/ruleset.xml
```diff
@@ -14,8 +14,8 @@
 		<type>error</type>
 		<message>File system writes only work in /tmp/ and inside the /uploads/ folder on VIP Go. To do filesystem writes you must use the WP_Filesystem class, using functions such as %s() won't work or will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
 	</rule>
-	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite">
-		<severity>6</severity>
+	<rule ref="WordPress.WP.AlternativeFunctions.file_system_read_file_put_contents">
+		<type>error</type>
```
GaryJones: Since this ruleset inherits the `WordPressVIPMinimum` one, where this overriding error is set (along with the other function), then it's not needed here as well.

rebeccahum:
Thanks, those are all valid points.  I suppose for our maintenance, it would be easier to just silence them on the WPCS side, as there are duplicate error messages.  In the `WordPress-VIP-Go` ruleset, we should probably change the error message to `File system writes only work in /tmp/ and inside the /uploads/ folder on VIP Go. To do filesystem writes you must use the WP_Filesystem class, using functions such as %s() won't work or will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/` instead of just `fwrite()` and `file_put_contents()` for consistency and less confusion methinks.

WordPress-VIP-Go/ruleset.xml
```diff
@@ -10,14 +10,70 @@
...
 	</rule>
 	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_fwrite">
 		<severity>6</severity>
 		<message>File system writes only work in /tmp/ and inside the /uploads/ folder on VIP Go. To do filesystem writes you must use the WP_Filesystem class, using functions such as %s() won't work or will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
 	</rule>
+	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_ftruncate">
+		<severity>6</severity>
+		<message>File system writes only work in /tmp/ and inside the /uploads/ folder on VIP Go. To do filesystem writes you must use the WP_Filesystem class, using functions such as %s() won't work or will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
+	</rule>
+	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_ftruncate">
```
GaryJones: This is the same as the rule above.

WordPressVIPMinimum/Sniffs/UserExperience/AdminBarRemovalSniff.php
```diff
@@ -411,4 +411,32 @@ protected function validate_css_property_value( $value, $compare_type, $compare_
 		}
 	}
 
+	/**
```
GaryJones: This change seems unrelated? Why is it in this PR?

GaryJones:
Audit has been updated.

PR #363, opened 2019-01-29
WIP Markdown Docs

tomjn:
Adds a subfolder with markdown docs


tomjn:
Once these are merged, the sniff messages themselves can be updated with links to the `master` branch copies of these files

PR #367, opened 2019-02-06
Added integration tests for WordPress-VIP-Go ruleset

rebeccahum:
This PR fixes #336 and also:

- Moves WordPressVIPMinimum integration test with updated references
- Merges `WordPressVIPMinimum.Security.PHPFilterFunctions.MissingSecondParameter`, `WordPressVIPMinimum.Security.PHPFilterFunctions.MissingThirdParameter` and  `WordPressVIPMinimum.Security.PHPFilterFunctions.RestrictedFilter` into parent rule `WordPressVIPMinimum.Security.PHPFilterFunctions` in WordPress-VIP-Go ruleset

PR #371, opened 2019-02-09
Ruleset Tests Improvements

GaryJones:
Follows on from #367, not that we've got two ruleset tests.

- Move mostly duplicate PHPCS_Ruleset_Test classes into new RulesetTest class.
- Refactor new class:
  - Accept a ruleset name in the constructor
  - Change public method from `run()` to `passes()`.
  - Break out logic into smaller private methods to make the logic more self-documenting.
  - Refactor variable names in some methods.
  - Decode JSON into objects, not arrays
  - Fix incorrect reference to local `$expected` to `$this->expected`. Somehow, this was still working regardless.
  - Add further documentation.
- Change ruleset test class usage, including adding the name to the "tests passed!" message.
- Replaced WPCS whitelisting `// XSS OK` comments in this files with `// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped` comments.
- Change naming of tests from "integration test" to "ruleset tests", to make it more intuitive exactly what these are (composer script, Travis, bin filename).

This also includes a fix for Travis, since it wasn't correctly allowing the PHP nightly build to fail.


WordPress-VIP-Go/ruleset-test.php
```diff
@@ -15,7 +15,7 @@
  * @package VIPCS\WordPressVIPGo
  */
 
-namespace WordPressVIPGo;
+namespace WordPressVIPMinimum;
```
GaryJones: While this file is in the WordPress-VIP-Go directory, I don't think there's any benefit in having a different namespace for this, at least at this stage. It can be renamed later on, if needed, since no-one else should be referencing anything inside this file.

PR #372, opened 2019-02-12
RestrictedFunctionsSniff: Add add_site_option function to site_option

### LINKED ISSUE #370: add_site_option in RestrictedFunctions grouping

Add `add_site_option()` to the RestrictedFunctions sniff, as we already have `update_site_option()` and `delete_site_option()` for consistent grouping.  We'll also need to silence it on the WordPress-VIP-Go ruleset.

```php
add_site_option( 'blahblah' );
```

```php
add_option();
```

rebeccahum:
Fixes #370.


rebeccahum:
I don't think so because that grouping of rules isn't explicitly listed in the `WordPressVIPMinimum` ruleset.

GaryJones:
> I don't think so because that grouping of rules isn't explicitly listed in the `WordPressVIPMinimum` ruleset.

They may be implicitly included by virtue of the whole directory of sniffs being included, but that doesn't mean we shouldn't be testing for them. There are plenty of `// WordPressVIPMinimum.` instances in the ruleset test, including a [whole block](https://github.com/Automattic/VIP-Coding-Standards/blob/master/WordPressVIPMinimum/ruleset-test.inc#L120) for RestrictedFunctions; though it looks incomplete, any new additions should be added anyway IMO.

rebeccahum:
If we are going to test for them, we should test for all being implicitly included and re-work the WordPressVIPMinimum integration tests to account for that in a different PR.

GaryJones:
> If we are going to test for them, we should test for all being implicitly included and re-work the WordPressVIPMinimum integration tests to account for that in a different PR.

Right - the ruleset test needs updating in another PR, but the bit relevant to this PR should be added in this PR - think atomic commits / merges.

rebeccahum:
Fair enough, done!

PR #377, opened 2019-02-13
Refactor to use VIPCSSniff which is child of WordPress\Sniff

### LINKED ISSUE #359: Consolidate common methods into base / abstract classes
We already have one abstract class, but there are other common idioms (like getting tokens from the phpcsFile, and often store them in a class property) and methods (multiple files deal with processing hooks, for example).

We should consider consolidating these into a base `Sniff` class, much like [WPCS does](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/develop/WordPress/Sniff.php), and making our sniff classes extends from this.

rebeccahum:
Fixes #359.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
@@ -9,7 +9,7 @@
 
 namespace WordPressVIPMinimum\Sniffs;
 
-use WordPress\Sniff;
+use WordPressVIPMinimum\Sniffs\Sniff;
 
 /**
  * Restricts usage of some variables.
--- a/WordPressVIPMinimum/Sniffs/Sniff.php
+++ b/WordPressVIPMinimum/Sniffs/Sniff.php
@@ -1,0 +1,20 @@
+<?php
+/**
+ * WordPressVIPMinimum Coding Standard.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ * @link https://github.com/Automattic/VIP-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressVIPMinimum\Sniffs;
+
+/**
+ * Represents a WordPress\Sniff for sniffing VIP coding standards.
+ *
+ * Provides a bootstrap for the sniffs, to reduce code duplication.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
+abstract class Sniff extends \WordPress\Sniff {
+}
--- a/WordPressVIPMinimum/Sniffs/Compatibility/ZoninatorSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Compatibility/ZoninatorSniff.php
@@ -7,15 +7,14 @@
 
 namespace WordPressVIPMinimum\Sniffs\Compatibility;
 
-use PHP_CodeSniffer\Files\File;
-use PHP_CodeSniffer\Sniffs\Sniff;
+use WordPressVIPMinimum\Sniffs\Sniff;
 use PHP_CodeSniffer\Util\Tokens;
 
 /**
  * This sniff reminds the developers to check whether the WordPress Core REST API is enabled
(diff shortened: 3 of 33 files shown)


WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
```diff
@@ -18,7 +18,7 @@
  *
  * @package VIPCS\WordPressVIPMinimum
  */
-abstract class AbstractVariableRestrictionsSniff extends Sniff {
+abstract class AbstractVariableRestrictionsSniff extends VIPCSSniff {
```
GaryJones: Instead of using a class with a different name, we should be able to remove the `use WordPress\Sniff;` and `use PHP_CodeSniffer\Sniffs\Sniff` import statements from each file that uses it, and keep `VIPCSSniff` as `Sniff`.

WordPressVIPMinimum/Sniffs/VIPCSSniff.php
```diff
@@ -0,0 +1,23 @@
+<?php
+/**
+ * WordPressVIPMinimum Coding Standard.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ * @link https://github.com/Automattic/VIP-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressVIPMinimum\Sniffs;
```
GaryJones: I'd suggest changing this to be:

```php
namespace WordPressVIPMinimum\Sniffs;

use WordPress\Sniff as WPCSSniff;

/**
 * Represents a WordPress\Sniff for sniffing VIP coding standards.
 *
 * Provides a bootstrap for the sniffs, to reduce code duplication.
 *
 */
abstract class Sniff extends WPCSSniff {
}
```

Or

```php
namespace WordPressVIPMinimum\Sniffs;

/**
 * Represents a WordPress\Sniff for sniffing VIP coding standards.
 *
 * Provides a bootstrap for the sniffs, to reduce code duplication.
 *
 */
abstract class Sniff extends WordPress\Sniff {
}
```

Then rename the file, and remove the WPCS & PHPCS `...\Sniff` import statements.

WordPressVIPMinimum/Sniffs/Compatibility/RobotstxtSniff.php
```diff
@@ -31,22 +31,19 @@ public function register() {
 	/**
 	 * Processes the tokens that this sniff is interested in.
 	 *
-	 * @param File $phpcsFile The PHP_CodeSniffer file where the token was found.
-	 * @param int  $stackPtr  The position in the stack where the token was found.
+	 * @param int $stackPtr  The position in the stack where the token was found.
```
GaryJones: There are still two spaces before the description (for each of these instances), which `WordPress-Docs` may complain about.

rebeccahum:
Thanks for the feedback, @GaryJones.  I think they should be all addressed now!

GaryJones:
Has a merge conflict.

WordPressVIPMinimum/Sniffs/Sniff.php
```diff
@@ -0,0 +1,22 @@
...
+ * WordPressVIPMinimum Coding Standard.
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ * @link https://github.com/Automattic/VIP-Coding-Standards
+ * @license https://opensource.org/licenses/MIT MIT
+ */
+
+namespace WordPressVIPMinimum\Sniffs;
+
+use WordPress\PHPCSHelper;
```
GaryJones: Why is this bit needed?

GaryJones:
I've:

 - Fixed the merge conflict with `RobotstxtSniff.php` by rebasing on the `master`.
 - Removed the unused import.
 - Split out the adding of the new sniff, and added more description to the commit message as to why an empty abstract class has been added.
 - Updated the commit message on the remaining commit to add a more detailed description about why so many `$phpcsFile` and `$token` variables have been updated to be references to the (grand-parent) class properties.

PR #382, opened 2019-02-15
Travis: Use "7.4snapshot" instead of nightly

GaryJones:
`nightly` means the PHP 8 dev branch, which phpcodesniffer-composer-installer does not support yet, but we can force a check on the 7.4 development branch with "7.4snapshot".


GaryJones:
Closing in favour of #383.

PR #386, opened 2019-02-21
WIP: Silence Short Echo Tag on VIP Go

GaryJones:
Short echo tags `<?=` has been constantly on since PHP 5.4. There is no fundamental reason why these should be disallowed.

This still forbids them for `WordPressVIPMinimum`, and it still forbids short open tags `<?` for both platforms.

I'd be open in removing them for `WordPressVIPMinimum` as well if needed.

WIP, since it needs ruleset tests to be updated.

### DIFF
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -283,5 +283,8 @@
 	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.site_option_add_site_option">
 		<severity>0</severity>
 	</rule>
+	<rule ref="Generic.PHP.DisallowShortOpenTag.EchoFound">
+		<severity>0</severity>
+	</rule>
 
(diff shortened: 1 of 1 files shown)


jenkoian:
Yeah I'd say when using PHP for templating (which is common in WordPress) supporting short echo tags makes a lot of sense.

GaryJones:
In this case, I noticed that MultilingualPress uses it a fair bit (350+ instances)!

GaryJones:
> WIP, since it needs ruleset tests to be updated.

This wasn't ready to merge yet!

rebeccahum:
@GaryJones Sorry, I must have missed what WIP means!

GaryJones:
> I must have missed what WIP means!

Work In Progress :-)

GaryJones:
@rebeccahum Did your recent ruleset test improvements catch the one missing for here?

Issue #375, opened 2019-02-12
Re-work WordPressVIPMinimum ruleset test 


rebeccahum:
## What problem would the enhancement address for VIP?
As discussed in #372, we should test for all WordPressVIPMinimum rules in our integration tests where they are being implicitly in the ruleset.  Currently, the tests only account for some rules (e.g. only some in the `RestrictedFunctionsSniff`).

rebeccahum:
We will need to figure out #395 to complete this.

GaryJones:
@rebeccahum Other than the JS tokenizer tests issue, is this otherwise now done?

rebeccahum:
@GaryJones Yep.

PR #400, opened 2019-03-28
Fix WordPress-VIP-Go integration tests

rebeccahum:
Since #397, it was unearthed that the number of errors/warnings on a line-basis were not being correctly found.  This PR adjusts the ruleset tests and also addresses most of #375.


WordPress-VIP-Go/ruleset-test.inc
```diff
@@ -322,20 +322,4 @@ update_site_option( $bar, $foo, true ); // Ok.
 // WordPressVIPMinimum.Functions.RestrictedFunctions.site_option_add_site_option
 add_site_option( 'foo', $bar ); // Ok.
 
-// WordPressVIPMinimum.Security.EscapingVoidReturnFunctions.Found
```
GaryJones: The point of the ruleset test, is to check that all of the sniffs a standard uses, trigger a violation (once of each type is enough), to prove that they are not accidentally silencing a violation, or that the other standard hasn't removed a violation without us realising.

IMO we should keep at least one bit of code that will trigger each `WordPressVIPMinimum` violation that `WordPress-VIP-Go` is consuming.

rebeccahum: Done.

PR #415, opened 2019-05-15
Downgrade file ops functions to warning

hanifn:
With stream wrappers it’s no longer an error


WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
```diff
@@ -219,16 +219,18 @@ public function getGroups() {
 					'session_write_close',
 				],
 			],
+			// @link VIP Go: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/
```
GaryJones: The VIP website domain has changed ;-)

GaryJones: We should update all of these references.

GaryJones:
@hanifn Thanks.

Can you please give example snippets of what code should be flagged, and which code shouldn't?

What I'm looking for is a way that someone could be using, `fclose()`, say:
- correctly, that we don't need to flag
- inconclusively, that we flag as a warning
- incorrectly, in a way that won't work, and should be flagged as an error.

GaryJones:
The ruleset tests will also need updating ([#](https://travis-ci.org/Automattic/VIP-Coding-Standards/jobs/532716885)), but I'd like to explore whether we can do a better job of detection first.

hanifn:
> What I'm looking for is a way that someone could be using, fclose(), say:

There are some functions like `fclose()` and `fwrite()` etc that are little tricky. Using it correctly or incorrectly is really dependent on how they created the file resource using `fopen()`.

For example:
```
// Correctly opening file
$basedir = wp_upload_dir()[ 'basedir' ]; // this will return the proper vip protocol stream
$file = fopen( $basedir . '/test.txt', '+r' );

// Incorrectly opening file
$file = fopen( './test.txt', '+r' );

// Thus the following functions would work or not depending on the how it's opened above
fread( $file, 1024 );
fwrite( $file, 'new string' );
fclose( $file );
```

GaryJones:
Does that apply to all of the items in the group?

Would it make more sense to create a sniff that focuses on the correct use of `fopen()`, if that's the starting point for all of them?

WordPress-VIP-Go/ruleset.xml
```diff
@@ -89,20 +89,24 @@
...
 	</rule>
 	<rule ref="WordPressVIPMinimum.JS.HTMLExecutingFunctions.append">
 		<type>error</type>
 	</rule>
 	<rule ref="WordPressVIPMinimum.Performance.FetchingRemoteData.FileGetContentsUnknown">
-		<type>error</type>
-		<message>%s() is uncached. If this is being used to query a remote file please use wpcom_vip_file_get_contents() instead. If it's used for a local file please use WP_Filesystem instead. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
+		<type>warning</type>
+		<message>%s() is uncached. File operations only work using the VIP stream on VIP Go. Make sure you are getting the proper base directory using wp_upload_dir() or it will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
```
mjangda: I think this one probably needs a bit more refinement since it changes the original intent of the error. There are three cases:

- local file
- file in the media library (via stream)
- remote file

Interestingly, `wpcom_vip_file_get_contents` does not work with a `vip://` stream URL, so we should revert back to the original message but with a slight tweak:

```suggestion
		<message>%s() is uncached. If the function is being used to fetch a remote file (e.g. a URL starting with https://), please use wpcom_vip_file_get_contents() to ensure the results are cached. For more details, please see https://wpvip.com/documentation/vip-go/fetching-remote-data/</message>
```

mjangda:
> Would it make more sense to create a sniff that focuses on the correct use of fopen(), if that's the starting point for all of them?

That sounds like a reasonable idea. Ideally, it we'd want look at any functions that that write based on a file path/stream path (e.g. `file_put_contents`) or any functions that return a file resource (`file`, `fopen`, etc.).

For a first pass at this, it would be nice to just downgrade the errors to warnings so we're no longer flagging legitimate use case and revisit more thorough checks in a separate PR.

WordPress-VIP-Go/ruleset.xml
```diff
@@ -9,63 +9,63 @@
 	<!-- Things that may be incompatible with the VIP Go infrastructure and needs a dev to review -->
 	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_delete">
 		<severity>6</severity>
-		<message>File system writes only work in /tmp/ and inside the /uploads/ folder on VIP Go. To do filesystem writes you must use the WP_Filesystem class, using functions such as %s() won't work or will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
+		<message>File system writes only work in /tmp/ or using the VIP stream on VIP Go. Make sure you are getting the proper base directory using wp_upload_dir() when using functions such as %s() or it will return unexpected results. Read more here: https://vip.wordpress.com/documentation/using-wp_filesystem-instead-of-direct-file-access-functions/</message>
```
mjangda: Here's a suggested re-wording of the message:

```suggestion
		<message>File system operations only work on the `/tmp/` and `wp-content/uploads/` directories. To avoid unexpected results, please use helper functions like `get_temp_dir()`  or `wp_get_upload_dir()` to get the proper directory path when using functions such as %s(). For more details, please see: https://wpvip.com/documentation/vip-go/writing-files-on-vip-go/</message>
```

We'll want to apply this to all the other ones as well.

mjangda:
@GaryJones thoughts on my suggestion?

> For a first pass at this, it would be nice to just downgrade the errors to warnings (with updated messaging) so we're no longer flagging legitimate use case and revisit more thorough checks in a separate PR.

WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
```diff
@@ -400,4 +402,6 @@ public function is_targetted_token( $stackPtr ) {
 		}
 		return false;
 	}
+
+
```
GaryJones: Remove these two clear linespaces please.

GaryJones:
Makes sense.

@hanifn Could you please rebase to fix the conflicts?

hanifn:
Sorry about that @GaryJones. Looks like i was working of `master` instead of `develop`. I've rebased and resolved the conflicts

GaryJones:
Noting that this will need our audit updating, since it changes the values in the Sniffs (not just the rulesets).

GaryJones:
Rebased on to `develop`, fixed merge conflicts, updated `ruleset-test.php` comments that identify where an Error or Warning is expected to be reported.

GaryJones:
@hanifn Just checking - you've made changes to the `RestrictedFunctionsSniff`, but this will also change how calls to the named functions will also be reported for WordPress.com code - was that intentional?

hanifn:
> was that intentional?

@GaryJones Definitely not intentional. How do I fix it? I'm not sure which part is used by WordPress.com code

GaryJones:
WordPress.com uses the sniffs, which may be overridden by the `WordPressVIPMinimum` ruleset.

VIP Go uses the `WordPressVIPMinimum` ruleset (which uses the sniffs), which are then overridden by the `WordPress-VIP-Go` ruleset.

As such, if you only want to affect behaviour for VIP Go, then make the changes in the `WordPress-VIP-Go` ruleset`.

(If you want to affect WordPress.com, make the changes in the sniffs.)

WordPressVIPMinimum/ruleset-test.php
```diff
@@ -262,6 +247,21 @@
 		288 => 1,
 		293 => 1,
 		294 => 1,
+		368 => 1,
```
GaryJones: Should the changes in this file be reverted? The change to Warnings should now only be for `WordPress-VIP-Go`, not `WordPressVIPMinimum`.

hanifn: Yes. Definitely will need to revert those. I'll work on that.

WordPress-VIP-Go/ruleset.xml
```diff
@@ -9,63 +9,63 @@
 	<!-- Things that may be incompatible with the VIP Go infrastructure and needs a dev to review -->
 	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_delete">
```
GaryJones: @hanifn I think each of the rules with amended messages here, need a `<type>warning</type>` element added.

GaryJones:
Excellent, thank you @hanifn!

PR #419, opened 2019-05-21
Update README.md

### LINKED ISSUE #301: WPCS 2.0: Bump PHPCS requirement to 3.3.1
VIPCS currently has PHPCS 3.2.3 as the minimum requirement.

[WPCS 2.0 will have PHPCS 3.3.1 as the minimum requirement](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/pull/1555). As part of VIPCS supporting WPCS 2.0, I propose that we also raise our minimum PHPCS requirement to 3.3.1.

As well as eliminating dependency differences between VIPCS and WPCS, we'll also benefit from having PHPCS fixes related to annotations, and PHP 7.3 support.

Our Travis matrix will also be simplified as well, since we won't have to explicitly exclude testing against PHPCS 3.2.3 against PHP 7.3+.

paulschreiber:
Set minimum requirement to 3.3.1. Fixes #301.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Alternatively, we recommend the [PHP_CodeSniffer Standards Composer Installer Pl
 
 ### Minimal requirements
 
-* [PHPCS 3+](https://github.com/squizlabs/PHP_CodeSniffer/releases)
+* [PHPCS 3.3.1+](https://github.com/squizlabs/PHP_CodeSniffer/releases)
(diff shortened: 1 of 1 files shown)


GaryJones:
Thanks Paul, but I'm going to close this in favour of https://github.com/Automattic/VIP-Coding-Standards/commit/d6d40c9ef74e54088d9f5d83262958915b339d6f (but thanks for the nudge!)

PR #437, opened 2019-08-18
HTMLExecutingFunctions: Add more functions

### LINKED ISSUE #435: Propably missing a warning on prepend method
as html , append here 

https://github.com/Automattic/VIP-Coding-Standards/blob/8544a94c32f990b6669f2c7f034d4ba75ccbb558/WordPressVIPMinimum/Sniffs/JS/HTMLExecutingFunctionsSniff.php#L27-L33

I have no warning when I ran WPCS VIP Minimum

GaryJones:
There are different ways to insert variables into the DOM, and our sniffs only covered some of them.

The sniff now has an expanded list, which covers functions that have a syntax where the content is in the function arg, and also where the function arg is the target and the content is in the preceding method's arg.

New functions:

- `after`
- `appendTo`
- `before`
- `insertAfter`
- `insertBefore`
- `prepend`
- `prependTo`
- `replaceAll`
- `replaceWith`

Fixes #435.

----

Not included is the ability to handle `insertAdjacentHTML()` or consolidating assignment to `innerHTML` (and `outerHTML`) into this sniff.


WordPressVIPMinimum/Sniffs/JS/HTMLExecutingFunctionsSniff.php
```diff
@@ -58,29 +71,56 @@ public function register() {
...
+		} elseif ( 'target' === $this->HTMLExecutingFunctions[ $this->tokens[ $stackPtr ]['content'] ] ) {
+			$prevToken = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $stackPtr - 1, null, true, null, true );
 
+			if ( T_OBJECT_OPERATOR !== $this->tokens[ $prevToken ]['code'] ) {
 				return;
 			}
+
+			$prevPrevToken = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $prevToken - 1, null, true, null, true );
+
+			if ( T_CLOSE_PARENTHESIS !== $this->tokens[ $prevPrevToken ]['code'] ) {
```
david-binda: I believe that we should also check whether the keyword we have found is a function call. As, for instance, the code, as it is now, would flag following code:

```
foo( variable ).appendTo;
```

WordPressVIPMinimum/Sniffs/JS/HTMLExecutingFunctionsSniff.php
```diff
@@ -58,29 +71,56 @@ public function register() {
...
+				return;
+			}
+
+			$parenthesis_opener = $this->tokens[ $prevPrevToken ]['parenthesis_opener'];
+
+			while ( $prevPrevToken > $parenthesis_opener ) {
+				$prevPrevToken = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $prevPrevToken - 1, null, true, null, true );
+				if ( T_STRING === $this->tokens[ $prevPrevToken ]['code'] ) { // Contains a variable.
```
david-binda: I know that this code has been in place prior this PR, but I don't think that the `// Contains a variable` inline comment is true, as even a function call would match the `T_STRING` code. Eg.:

```
$( foo() ).appendTo;
```

GaryJones:
Decided to prove to myself that these did indeed execute JavaScript: https://jsbin.com/pohefecucu/edit?html,js,output (creates 15 alerts that you'll need to click through!)

There are some subtleties:
 - using `append()` on a jQuery object will evaluate (run) what is added, but using `append()` with a JavaScript Node will not. There's simply no way to accurately determine what `append()` is being called on from the tokens, hence needing to treat them all the same.
 - Script elements inserted using `innerHTML` do not execute when they are inserted, but there are other ways to mis-use `innerHTML` and result in XSS. We already [check for the presence of a variable](https://github.com/Automattic/VIP-Coding-Standards/blob/develop/WordPressVIPMinimum/Sniffs/JS/InnerHTMLSniff.php#L68) in our innerHTML sniff.

PR #440, opened 2019-08-18
Redirect functions: check for third parameter

### LINKED ISSUE #433: Check for third arg of wp_redirect() / wp_safe_redirect()
Debugging redirects can be hard as there are so many levels where a redirect may happen. 

WordPress 5.1 and later adds an `X-Redirect-By` header when `wp_redirect()` or `wp_safe_redirect()` is used, with a default value of `WordPress`, but this can be customised to any string value by adding a third argument to those function calls. This would make debugging redirects easier.

Depending on whether we could make this mandatory, this would be an Error or Warning if the argument was missing. There would be no requirement on what the string should say, but the expectation is that it would be representative of the client's theme or plugin name.

```php
wp_safe_redirect( $location, 302 );
wp_redirect( 'https://...', 301 );
```

```php
wp_safe_redirect( $location, 302, 'Client Z Redirects plugin' );
wp_redirect( 'https://...', 301, 'Foo theme' );
```

GaryJones:
Debugging redirects can be hard as there are so many levels where a redirect may happen.

WordPress 5.1 and later adds an `X-Redirect-By` header when `wp_redirect()` or `wp_safe_redirect()` is used, with a default value of `WordPress`, but this can be customised to any string value by adding a third argument to those function calls. This would make debugging redirects easier.

This PR adds a Warning if the third arg is missing.

Fixes #433.


rebeccahum:
I like the idea and it would definitely be a better practice... but at the same time, I think this would add _a lot_ of noise since typically the third parameter isn't filled out as it's optional.

PR #444, opened 2019-08-27
ConstantStringSniff: Take into account namespace prefixing

### LINKED ISSUE #422: ConstantString doesn't allow for namespaces before the string
## Bug Description

[ConstantStringSniff](https://github.com/Automattic/VIP-Coding-Standards/blob/master/WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php) doesn't allow for namespaces before the string.

## Minimal Code Snippet

```php
namespace Foo\Bar;

const REST_ALLOWED_META_PREFIXES = [ 'foo-', 'bar-', 'baz-' ];

if ( defined( __NAMESPACE__ . '\REST_ALLOWED_META_PREFIXES' ) && in_array( 'foo-', REST_ALLOWED_META_PREFIXES, true ) ) {
}
```

## Error Code

> Constant name, as a string, should be used along with `defined()`.
> (WordPressVIPMinimum.Constants.ConstantString.NotCheckingConstantName)

## Environment

| PHP version             | 7.3.2 
| PHP_CodeSniffer version | 3.4.2
| VIPCS version           | 1.0.0

- [X] I have verified the issue still exists in the `master` branch of VIPCS.

rebeccahum:
Fixes #422.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php
@@ -55,7 +55,13 @@ class ConstantStringSniff extends Sniff {
 			return;
 		}
 
-		$nextToken = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextToken + 1, null, true, null, true );
+		$nextToken     = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextToken + 1, null, true, null, true );
+		$nextNextToken = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextToken + 1, null, true, null, true );
+
+		if ( T_NS_C === $this->tokens[ $nextToken ]['code'] && T_STRING_CONCAT === $this->tokens[ $nextNextToken ]['code'] ) {
+			// Namespacing being used, skip to next.
+			$nextToken = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextNextToken + 1, null, true, null, true );
+		}
 
 		if ( $this->tokens[ $nextToken ]['code'] !== T_CONSTANT_ENCAPSED_STRING ) {
 			$message = 'Constant name, as a string, should be used along with `%s()`.';
--- a/WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php
@@ -55,20 +55,26 @@ class ConstantStringSniff extends Sniff {
 			return;
 		}
 
-		$nextToken     = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextToken + 1, null, true, null, true );
-		$nextNextToken = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextToken + 1, null, true, null, true );
-
-		if ( T_NS_C === $this->tokens[ $nextToken ]['code'] && T_STRING_CONCAT === $this->tokens[ $nextNextToken ]['code'] ) {
-			// Namespacing being used, skip to next.
-			$nextToken = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextNextToken + 1, null, true, null, true );
-		}
-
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php
```diff
@@ -55,7 +55,13 @@ public function process_token( $stackPtr ) {
 			return;
 		}
 
-		$nextToken = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextToken + 1, null, true, null, true );
+		$nextToken     = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextToken + 1, null, true, null, true );
+		$nextNextToken = $this->phpcsFile->findNext( Tokens::$emptyTokens, $nextToken + 1, null, true, null, true );
+
+		if ( T_NS_C === $this->tokens[ $nextToken ]['code'] && T_STRING_CONCAT === $this->tokens[ $nextNextToken ]['code'] ) {
+			// Namespacing being used, skip to next.
```
GaryJones: Could [`is_token_namespaced()`](https://github.com/WordPress/WordPress-Coding-Standards/blob/06966b26d8deb06a91a5c0f99a04802a6ff5e5bf/WordPress/Sniff.php#L1648-L1671) be used here?

GaryJones: I think you're right. I had a quick look to see if there were any other utility functions in WPCS that might be able to check for this, but couldn't spot any.

@jrfnl - one for your new Utils package?

jrfnl: Possibly, but to be fair, when `__NAMESPACE__` is used, it doesn't necessarily mean that the code is namespaced. `__NAMESPACE__` can evaluate to an empty namespace.

If you like, you can open an issue already in the [Utils repo](https://github.com/PHPCSStandards/PHPCSUtils). The more detailed the issue, the better.

My initial priority for the Utils repo is to get it set up and release v 1.0 with everything from the refactor PR and a little more.

After that I will look at any open issues, so it may be a little while before I even look at it.

WordPressVIPMinimum/Tests/Constants/ConstantStringUnitTest.inc
```diff
@@ -4,6 +4,12 @@ if ( ! defined( 'WPCOM_VIP' ) ) { // Okay.
...
 
-if ( ! defined( WPCOM_VIP ) ) { // NOK.
-	define( WPCOM_VIP ); // NOK.
-}
\ No newline at end of file
+if ( ! defined( WPCOM_VIP ) ) { // Error.
+	define( WPCOM_VIP ); // Error.
+}
+
+namespace Foo/Bar;
```
jrfnl:
```suggestion
namespace Foo\Bar;
```

jrfnl:
I've added a new commit to this PR to implement option 1 from my comment in the original ticket.

## Commit details:

### Constants/ConstantString: prevent more false positives

The `define()` and `defined()` functions expect the name of a constant as a text string to be passed as the first parameter.
This sniff is intended to guard against a coding mistake where the name of the constant would be passed as a_constant_.

While passing the name of the constant as a constant _could_ be perfectly valid when that constant contains the name of another constant as a text string, more often than not, it will be a coding error.

As things were, the sniff did not handle parameters build up using more than one token into account. It also did not handle valid situations, like passing the constant name as a text string via a variable to the functions, correctly.

As outlined in https://github.com/Automattic/VIP-Coding-Standards/issues/422#issuecomment-698934220 (Proposal 1), this limits the sniff to _only_ trigger an error when a plain constant has been passed as the "constant name" parameter and ignore all other cases.

Note: The current implementation of this fix, uses the WPCS `get_function_call_parameter()` method to retrieve the stack pointer to the start and end of the parameter. Once PHPCSUtils is in place, this function call should be replaced by using the PHPCSUtils version of that method.

Fixes #422
Fixes #480

GaryJones (approved):

PR #445, opened 2019-08-29
LowExpiryCacheTimeSniff: Account for random generating number func

### LINKED ISSUE #408: Enhance logic of LowCacheTime to minimize noise.
`❗️ Warning: Low cache expiry time of "wp_rand( 5 * MINUTE_IN_SECONDS, 10 * MINUTE_IN_SECONDS )", it is recommended to have 300 seconds or more (WordPressVIPMinimum.Cache.LowExpiryCacheTime.LowCacheTime).`

`wp_cache_set( $latest_blocks_cache_key, $latest_blocks, null, wp_rand( 5 * MINUTE_IN_SECONDS, 10 * MINUTE_IN_SECONDS ) );`

It is very common for clients to use constants HOUR_IN_SECONDS, DAY_IN_SECONDS, MINUTE_IN_SECONDS and to also vary the duration to not have cache stampedes.

Can we enhance the logic of the ruleset to not flag for common usages?

rebeccahum:
Fixes #408

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Performance/LowExpiryCacheTimeSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Performance/LowExpiryCacheTimeSniff.php
@@ -54,6 +54,18 @@ class LowExpiryCacheTimeSniff extends AbstractFunctionParameterSniff {
 	];
 
 	/**
+	 * List of random generating number functions.
+	 *
+	 * @var array
+	 */
+	protected $rand_functions = [
+		'wp_rand',
+		'random_int',
+		'mt_rand',
+		'rand',
+	];
+
+	/**
 	 * Process the parameters of a matched function.
 	 *
 	 * @param int    $stackPtr        The position of the current token in the stack.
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Performance/LowExpiryCacheTimeSniff.php
```diff
@@ -69,21 +81,34 @@ public function process_parameters( $stackPtr, $group_name, $matched_content, $p
...
 			}
 		}
 
-		if ( $time < 300 ) {
-			$message = 'Low cache expiry time of "%s", it is recommended to have 300 seconds or more.';
-			$data    = [ $parameters[4]['raw'] ];
-			$this->phpcsFile->addWarning( $message, $stackPtr, 'LowCacheTime', $data );
+		$times = [];
+		if ( false !== $rand_function ) {
+			$times = explode( ',', preg_replace( '/[( )|\(|\)|(' . $rand_function . ')]/', '', $time ) );
```
GaryJones, line 97: I think this regex is going to miss cases where there are unusual code styling - line breaks before, between, or after string arguments like `20 * HOUR_IN_SECONDS` - I'd at least like to see more unit tests, and I think when we switch to be able to use https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Utils-PassedParameters.html#method_getParameters then this will be much easier to be confident with.

GaryJones (changes requested):

GaryJones:
Conflicts need resolving before it can be reviewed.

PR #450, opened 2019-10-31
Switch to sirbrillig/phpcs-variable-analysis

### LINKED ISSUE #449: Replace VariableAnalysis Sniff with VariableAnalysis package
Follow-on to #236.

We currently have a [VariableAnalysis sniff](https://github.com/Automattic/VIP-Coding-Standards/blob/0363baf7ae0d92b2662bc0140da12709cbc7ee02/WordPressVIPMinimum/Sniffs/Variables/VariableAnalysisSniff.php), but it has some issues. It currently provides the following violations:

- `WordPressVIPMinimum.Variables.VariableAnalysis.VariableRedeclaration`: Warning, severity 5, "Redeclaration of %s %s as %s."
- `WordPressVIPMinimum.Variables.VariableAnalysis.UndefinedVariable`: Warning, severity 5, "Variable %s is undefined."
- `WordPressVIPMinimum.Variables.VariableAnalysis.UnusedVariable`: Warning, severity 5, "Unused %s `%s`."
- `WordPressVIPMinimum.Variables.VariableAnalysis.SelfInsideClosure`: Error, severity 5, "Use of `{$err_desc}%s` inside closure."
- `WordPressVIPMinimum.Variables.VariableAnalysis.SelfOutsideClass`: Error, severity 5, "Use of `{$err_desc}%s` outside class definition."
- `WordPressVIPMinimum.Variables.VariableAnalysis.StaticInsideClosure`: Error, severity 5, "Use of `{$err_desc}%s` inside closure."
- `WordPressVIPMinimum.Variables.VariableAnalysis.StaticOutsideClass`: Error, severity 5, "Use of `{$err_desc}%s` outside class definition."

In #236, it was decided to replace this with https://github.com/sirbrillig/phpcs-variable-analysis (cc @sirbrillig) - this provides the following violations:

- `VariableAnalysis.CodeAnalysis.VariableAnalysis.VariableRedeclaration`: Warning, severity 5, "Redeclaration of %s %s as %s."
- `VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable`: Warning, severity 5, "Variable %s is undefined."
- `VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable`: Warning, severity 5, "Unused %s %s."
- `VariableAnalysis.CodeAnalysis.VariableAnalysis.SelfOutsideClass`: Error, severity 5, "Use of self::%s outside class definition."
- `VariableAnalysis.CodeAnalysis.VariableAnalysis.StaticOutsideClass`: Error, severity 5, "Use of static::%s outside class definition."

There is no `SelfInsideClosure` or `StaticInsideClosure` violations, as these appear to be allowed now in modern versions of PHP. The rest of the detection process is more maintained, and handles PHP 7 code better. Switching to it would remove some common false positives.

We could also take this opportunity is taken to silence the UnusedVariable violation. It's severity is already dropped down to 1 for `WordPress-VIP-Go`, which is the same as what the bot runs at (though this may change in the future). From a VIP perspective, unused variables are not important.

_Undefined_ variables, however, will be flagged in PHP 8, so we should maintain this violation (`WordPress-VIP-Go` sets it as severity 3), even if the message is changed to highlight this incompatibility with PHP 8.

## False Positives

The following snippet is often highlighted by the existing sniff as `$this` being undefined when in a closure in a class. That should no longer be flagged by the new package.

```php
class MyClass {
	function my_function() {
		return function() {
			$this->my_callback();
		};
	}

	function my_callback() {}
}
```

The old sniff should be kept and marked as deprecated, to avoid breaking compatibility with custom rulesets, until the next major release of VIPCS when it can be removed.

GaryJones:
Introduces a new dependency of `sirbrillig/phpcs-variable-analysis` which is a more maintained version of the existing `VariableAnalysis` sniff in VIPCS.

The new sniff is added to the `WordPressVIPMinimum` ruleset. `WordPress-VIP-Go` ruleset inherits this, and maintains its dropping of the severity for `UnusedVariable` and `UndefinedVariable` violations.

The ruleset tests have set the previously undefined variables (or changed some variable names), so that violations on each line are still specific to the thing being tested, and doesn't include extra violations due to the use of throwaway variables.

One of the reasons for changing was the previously false positive with respect to the use of `$this` inside closures inside class methods. A ruleset test has been added to cover this, and it correctly does not get flagged in the ruleset test.

The old `VariableAnalysis` has been marked as deprecated, until it can be removed in the next major release of VIPCS. This will stop any breakage of setups which reference `WordPressVIPMinimum.Variables.VariableAnalysis` in a custom ruleset. If called, it throws a warning, and then passes control to the new VariableAnalysis sniff. The deprecation notice is excluded if the sniff is not explicitly included, and the duplicate violation messages are also silenced.

The previously used `VariableAnalaysisHelper.php` has been removed completely, as no-one should have been referencing that directly.

Fixes #449.


WordPressVIPMinimum/ruleset.xml
```diff
@@ -157,10 +157,35 @@
...
-		</properties>
+	<!-- Include VariableAnalysis checks -->
+	<rule ref="VariableAnalysis"/>
+
+	<!-- Prevent deprecation notice when the sniff is not explicitly included. -->
+	<rule ref="WordPressVIPMinimum.Variables.VariableAnalysis.DeprecatedSniff">
+		<severity>0</severity>
+	</rule>
+
+	<!-- Prevent duplicate messages from deprecated sniff. -->
```
rebeccahum: I'm a bit confused, how come we want to silence these below warnings if they are being removed in the VariableAnalysisSniff?

GaryJones: This is how deprecations are done in WPCS. If I comment these out, I do get duplicate messages, as PHPCS has registered the same violations under two different sniff classes (albeit that both ultimately use the same logic from the new package class).

GaryJones:
> Is there a way to create an "alias", so this won't break the `// phpcs:ignore WordPressVIPMinimum.Variables.VariableAnalysis.UndefinedVariable` comments?

Not as far as I know. The deprecation handles references in custom rulesets, but not ignore / disable references.

GaryJones:
Noting that VA 2.6.3+ requires PHP 5.6, so a composer install may need to ignore platform requirements to be able to test PHP 5.4 and PHP 5.5.

GaryJones:
Refactored and fixed up tests and updated to use VA `^2.8.2`. Currently failing on PHP 5.4 and PHP 5.5 Travis jobs.

I tested ignoring platform requirements so that VA could be installed on PHP 5.4 and 5.5, but a) it fails due to the PHP 5.6+ code in VA, and b) ignoring platform requirements make no sense when this is a non-dev dependency.

I suspect my [comment](https://github.com/sirbrillig/phpcs-variable-analysis/pull/86#issuecomment-657076732) isn't going to result in VA moving back to PHP 5.4 due to the code in it, so we'll have to bump the minimum PHP for VIPCS back up (having only [just dropped it](https://github.com/Automattic/VIP-Coding-Standards/pull/491)) to PHP 5.6.

GaryJones:
Before merging, we should check the changes here against the proposals made in https://github.com/Automattic/VIP-Coding-Standards/issues/449, or at least open a follow-up PR which the extra changes, within the same release.

sirbrillig:
> I suspect my comment isn't going to result in VA moving back to PHP 5.4 due to the code in it, so we'll have to bump the minimum PHP for VIPCS back up (having only just dropped it) to PHP 5.6.

Surprise! 😁  https://github.com/sirbrillig/phpcs-variable-analysis/pull/191

I'll still need to backport it to the 2.x branch though. Coming soon...

sirbrillig:
https://github.com/sirbrillig/phpcs-variable-analysis/releases/tag/v2.8.3 👍

GaryJones:
@sirbrillig You are a wonderful human - thank you!

PR #455, opened 2019-12-09
Travis: test against PHP 7.4, not snapshot

GaryJones:
Also test against a `nightly` (PHP 8.0), which is allowed to fail.


rebeccahum:
@GaryJones I was able to reproduce the failed tests with Composer 1.9.1 + PHP 7.4.  Can we bump to use Composer 1.9.2 for 7.4?

PR #464, opened 2020-01-31
RestrictedFunctions: Add stats_get_csv()

### LINKED ISSUE #461: Add stats_get_csv to restricted functions

Every time `stats_get_csv()` is called, it stores the data in the option `stats_cache`, which is where it can be polluted, as Jetpack relies on this value: https://github.com/Automattic/jetpack/blob/29b6f22dd57d0a5e8967fef73279dfdc575b820d/modules/stats.php#L1490. 

This function should not be relied on and a custom solution (e.g. stats endpoint) should be used instead.

```
$popular = stats_get_csv( 'postviews', [ 'days'  => 2,   'limit' => 20 ] );
```

```
$popular = custom_stats_get_csv( 'postviews', [ 'days'  => 2,   'limit' => 20 ] );
```

rebeccahum:
Fixes #461.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
@@ -31,7 +31,6 @@ class IncludingFileSniff extends AbstractFunctionRestrictionsSniff {
 		'get_stylesheet_directory',
 		'get_template_directory',
 		'locate_template',
-		'get_parent_theme_file_path'
 	];
 
 	/**
--- a/WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
@@ -268,6 +268,13 @@ class RestrictedFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 					'delete_site_option',
 				],
 			],
+			'stats_get_csv' => [
+				'type'      => 'error',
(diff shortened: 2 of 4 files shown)


WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
```diff
@@ -268,6 +268,13 @@ public function getGroups() {
 					'delete_site_option',
 				],
 			],
+			'stats_get_csv' => [
+				'type'      => 'error',
+				'message'   => 'Using `%s` outside of Jetpack context pollutes the stats_cache option. We recommend building a custom function instead of relying on it.',
```
GaryJones:
```suggestion
				'message'   => 'Using `%s` outside of Jetpack context pollutes the stats_cache entry in the wp_options table. We recommend building a custom function instead.',
```

GaryJones:
Needs unit tests.

rebeccahum:
@GaryJones Whoops. Done!

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -31,7 +31,6 @@ class IncludingFileSniff extends AbstractFunctionRestrictionsSniff {
 		'get_stylesheet_directory',
 		'get_template_directory',
 		'locate_template',
-		'get_parent_theme_file_path'
```
GaryJones: Think this came across from another PR?

PR #485, opened 2020-07-03
Ruleset Tests: Avoid issue with trigger_error()

GaryJones:
For some reason, the presence of `trigger_error()` in our ruleset tests (either directly, or via the presence of `_deprecated_file()`, was breaking the tests.

This should probably be looked at again to find the root cause, but this is a satisfactory work around for now.


GaryJones:
Judging between how https://travis-ci.org/github/Automattic/VIP-Coding-Standards/jobs/704661875 passed (for a different PR), and https://travis-ci.org/github/Automattic/VIP-Coding-Standards/jobs/704661876 failed (same, but with PHP 7.4snapshot), it looks like something changed with `trigger_error()` + PHP 7.4.

GaryJones:
Once #484 is merged, then the final job should be re-run, as the curly brace in Config.php error was [fixed in PHPCS 3.5.0](https://github.com/squizlabs/PHP_CodeSniffer/issues/2558)

GaryJones:
Reverted and fixed in #509.

PR #560, opened 2020-07-27
Allow for checking test code coverage

jrfnl:
As part of the sniff reviews, I'll be checking that the sniffs are properly covered by unit tests.

Adding a code coverage configuration and `@covers` tags makes that simpler.

**Note**: this doesn't yet add code coverage checking to the Travis build/PR build checks. If so desired, that can be added separately.

The current code coverage results as is:
image

###  Tests: add code coverage configuration

* PHPUnit config: add code coverage configuration.
    By default, when there is a code coverage configuration and Xdebug is enabled, code coverage will be generated.
    Note: generating code coverage is slow.
* Git ignore the `build` directory as created by PHPUnit to store the log files.
* Adjust the "normal" test script to not generate code coverage information.
* Add a `bin/unit-tests-coverage` to run the unit tests with code coverage enabled.
* Add a Composer script to call the `unit-tests-coverage` script.

### Code coverage recording: add @Covers tags

... to all unit test files, as well as enable strict coverage recording.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+build/
 vendor/
 composer.lock
 phpcs.xml
--- a/WordPressVIPMinimum/Tests/Classes/DeclarationCompatibilityUnitTest.php
+++ b/WordPressVIPMinimum/Tests/Classes/DeclarationCompatibilityUnitTest.php
@@ -13,6 +13,8 @@ use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
  * Unit test class for the DeclarationCompatibility sniff.
  *
  * @package VIPCS\WordPressVIPMinimum
+ *
+ * @covers \WordPressVIPMinimum\Sniffs\Classes\DeclarationCompatibilitySniff
  */
 class DeclarationCompatibilityUnitTest extends AbstractSniffUnitTest {
 
--- a/WordPressVIPMinimum/Tests/Classes/RestrictedExtendClassesUnitTest.php
+++ b/WordPressVIPMinimum/Tests/Classes/RestrictedExtendClassesUnitTest.php
@@ -13,6 +13,8 @@ use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
  * Unit test class for the RestrictedExtendClasses sniff.
  *
  * @package VIPCS\WordPressVIPMinimum
+ *
+ * @covers \WordPressVIPMinimum\Sniffs\Classes\RestrictedExtendClassesSniff
  */
 class RestrictedExtendClassesUnitTest extends AbstractSniffUnitTest {
 
--- a/WordPressVIPMinimum/Tests/Compatibility/ZoninatorUnitTest.php
+++ b/WordPressVIPMinimum/Tests/Compatibility/ZoninatorUnitTest.php
@@ -13,6 +13,8 @@ use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
  * Unit test class for the CheckReturnValue sniff.
  *
  * @package VIPCS\WordPressVIPMinimum
+ *
+ * @covers \WordPressVIPMinimum\Sniffs\Compatibility\ZoninatorSniff
  */
 class ZoninatorUnitTest extends AbstractSniffUnitTest {
 
--- a/WordPressVIPMinimum/Tests/Constants/ConstantStringUnitTest.php
+++ b/WordPressVIPMinimum/Tests/Constants/ConstantStringUnitTest.php
@@ -13,6 +13,8 @@ use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
  * Unit test class for the ConstantString sniff.
  *
  * @package VIPCS\WordPressVIPMinimum
+ *
+ * @covers \WordPressVIPMinimum\Sniffs\Constants\ConstantStringSniff
  */
 class ConstantStringUnitTest extends AbstractSniffUnitTest {
 
--- a/WordPressVIPMinimum/Tests/Constants/RestrictedConstantsUnitTest.php
+++ b/WordPressVIPMinimum/Tests/Constants/RestrictedConstantsUnitTest.php
@@ -13,6 +13,8 @@ use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
  * Unit test class for the ConstantRestrictions sniff.
  *
  * @package VIPCS\WordPressVIPMinimum
+ *
+ * @covers \WordPressVIPMinimum\Sniffs\Constants\RestrictedConstantsSniff
  */
 class RestrictedConstantsUnitTest extends AbstractSniffUnitTest {
 
--- a/WordPressVIPMinimum/Tests/Files/IncludingFileUnitTest.php
+++ b/WordPressVIPMinimum/Tests/Files/IncludingFileUnitTest.php
@@ -12,6 +12,8 @@ use PHP_CodeSniffer\Tests\Standards\AbstractSniffUnitTest;
  * Unit test class for the IncludingFile sniff.
  *
  * @package VIPCS\WordPressVIPMinimum
+ *
+ * @covers \WordPressVIPMinimum\Sniffs\Files\IncludingFileSniff
  */
 class IncludingFileUnitTest extends AbstractSniffUnitTest {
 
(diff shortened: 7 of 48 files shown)


bin/unit-tests-coverage.txt
```diff
@@ -0,0 +1,16 @@
...
+# EXAMPLE TO RUN LOCALLY:
+#
+#   ./bin/unit-test-coverage
+#
+# If you want to view the results locally, run with the following extra options:
+#
+#   ./bin/unit-test-coverage --coverage-html ./build/logs/
+#
+
+"$(pwd)/vendor/bin/phpunit" --filter WordPressVIPMinimum "$(pwd)/vendor/squizlabs/php_codesniffer/tests/AllTests.php" %*
```
GaryJones, line 16: What does the `%*` do here?

jrfnl: It will take any additional arguments you pass on the command line and append them.

Without it, the second example of how to run the command (line 13) wouldn't (shouldn't) work as `--coverage-html ./build/logs/` would be thrown away instead of added to the command.

Hmm.. trying to remember now whether this is a Windows syntax or cross-platform. Would you mind testing both the commands ? IIRC you are on Mac or not ?

jrfnl:
I probably should have mentioned in the issue above that generating code coverage only works with Xdebug turned on. If it's turned off, the tests will still run, but no coverage reports are generated.

jrfnl:
@GaryJones And you've just demonstrated why I find it important that people actually test my PRs ;-) Thanks a lot for this!

* `txt` extension removed (shouldn't have been there in the first place).
* `%*` removed. Windows compatibility for the script is irrelevant as these type of scripts don't run on Windows anyway, so I got my own `.bat` files set up.
* Typo in `composer.json` fixed.

Remaining:
* Making the file executable (chmod). That's not something I can do on Windows, so would be great if someone else could add/adjust a commit to make the file executable.
* Testing the command for taking additional arguments as per the second example `./bin/unit-test-coverage --coverage-html ./build/logs/`.
    When run like that, the code coverage report is generated in HTML as per the screenshot above, which is, of course, a lot more user friendly than a Clover XML report.
    If this doesn't work out of the box, the command may need to be adjusted still - I have a niggling feeling `$*` might be the equivalent of the previous `%*` for *nix based OS-es.
    If the ability to pass additional arguments is not needed, then the second example in the script file should be removed.

GaryJones:
> Making the file executable (chmod). That's not something I can do on Windows, so would be great if someone else could add/adjust a commit to make the file executable.

Done. Squashed into the commit that added the file in the first place (first commit).

> Testing the command for taking additional arguments as per the second example `./bin/unit-test-coverage --coverage-html ./build/logs/`.

As noted previously, the `%*` gave me an error. `$*` and `$@` both worked, though I suspect they wouldn't work for Windows users. But...

> If the ability to pass additional arguments is not needed, then the second example in the script file should be removed.

Since generation is slow, then this is likely to be done as a manual (not CI) task, so if a `clover.xml` is going to be generated, then we may as well generate the HTML output as well. I swapped out the `%*` for `--coverage-html ./build/logs/`, so now calling `composer coverage` generates the HTML.
(Squashed into the same first commit.)

jrfnl:
@GaryJones Thank you very much for making those changes.

>  if a clover.xml is going to be generated, then we may as well generate the HTML output as well. I swapped out the %* for --coverage-html ./build/logs/, so now calling composer coverage generates the HTML.

For now, the clover file would then not be needed and the config to generate the HTML output could be added to the PHPUnit config file instead (and removed from the script).

Would you like me to change that ?

Other than that - the ability to pass extra command line arguments to the script can still be useful, both for the test as well as the coverage script.

As previously said, it's not relevant for me as I run these via my own bat scripts anyway, but as an example of typical other command line args I use:
* `bin/unit-tests --filter AlwaysReturnInFilter` just runs the tests for that one sniff instead of all tests which can be useful while working on one particular sniff.


> Since generation is slow, then this is likely to be done as a manual (not CI) task

I imagine it can be added to CI at some point, but yes, that's not high priority at the moment.

For reference - both PHPCompatibility and PHPCSUtils check code coverage during CI and as part of the required status checks for each PR.

The way this is set up:
* Register with an external service which tracks code coverage across builds, The above both use Coveralls, but there are other services which can do this too.
* For Coveralls, a config file is needed.
* Then in Travis CI, we have slightly different set up:
    - For non-PR/merge builds we don't run the tests against all versions, but do a quick test against just the high/low PHP and high/low PHPCS versions (like in WPCS)
    - For PR/merge builds, we run the tests against all other PHP/PHPCS version combi's and run just the high/low combis from the quicktest stage with code coverage.

As the high/low PHPCS versions for VIPCS are so close, I can even imagine that for VIPCS a coverage stage with only two builds, PHP 7.4 + PHPCS `master` and PHP 5.4 + PHPCS "low" would be sufficient.

GaryJones:
> For now, the clover file would then not be needed and the config to generate the HTML output could be added to the PHPUnit config file instead (and removed from the script).
>
> Would you like me to change that ?

Yes please.

> As previously said, it's not relevant for me as I run these via my own bat scripts anyway, but as an example of typical other command line args I use:
>
> - bin/unit-tests --filter AlwaysReturnInFilter just runs the tests for that one sniff instead of all tests which can be useful while working on one particular sniff.

In which case, please go ahead and restore `$*` (although one page I was reading yesterday suggested that `$@` was the safer option), and include your example of being able to run tests for a single sniff in CONTRIBUTING.md / the shell script(s), as that would be useful to have for the future.

jrfnl:
Done - I've made the base config + script change in the first commit and added a third commit adding the ability to pass additional arguments to both scripts.

jrfnl:
Re: Contributing.md - I've not made any changes there (yet) as it looks like that whole section should probably get a rewrite. Currently looks to be taken largely from WPCS and doesn't account for the scripts used in this repo yet,

Might be more efficient to that rewrite when PHPCSUtils is added, as that section would need to change then anyway.

jrfnl:
@GaryJones Argh... ok, did some digging and the `processUncoveredFilesFromWhitelist="true"` was the cause as it loads the classes prior to running the tests which interferes with the PHPCS native autoloader.

I've removed that now from the first commit and :crossed_fingers: all should work fine again now, including generating the code coverage.

GaryJones:
✅

```sh
./bin/unit-tests-coverage --filter AlwaysReturnInFilter
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

.                                                                   1 / 1 (100%)

1 sniff test files generated 2 unique error codes; 0 were fixable (0%)

Time: 945 ms, Memory: 14.00 MB
```

❌ 

```sh
composer coverage
> bin/unit-tests-coverage
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

...........................................                       43 / 43 (100%)

44 sniff test files generated 120 unique error codes; 0 were fixable (0%)

Time: 6.48 seconds, Memory: 18.00 MB

OK (43 tests, 0 assertions)

Generating code coverage report in HTML format ... failed
Directory "/build/logs/" does not exist.
```

Not sure what's changed here, as I didn't have the `/build/logs/` before, and they were created just fine previously.

jrfnl:
> Not sure what's changed here, as I didn't have the /build/logs/ before, and they were created just fine previously.

PHPUnit creates that fine for me. Could it be that there is still a lock on the name on your system as it previously existed ?
Have you tried manually creating the directory and running the command with the (empty) directory in place ?

rebeccahum:
I get the same error as well even with manually running the command with the empty directory in place:

```
> bin/unit-tests-coverage
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

...........................................                       43 / 43 (100%)

44 sniff test files generated 120 unique error codes; 0 were fixable (0%)

Time: 6.85 seconds, Memory: 18.00 MB

OK (43 tests, 0 assertions)

Generating code coverage report in HTML format ... failed
Directory "/build/logs/" does not exist.
```

jrfnl:
Hmm... just to see if it makes a difference, could you try changing the `<log type="coverage-html" target="/build/logs/"/>` to `<log type="coverage-html" target="/build/logs"/>` ? i.e. without the trailing slash in the path.

rebeccahum:
`<log type="coverage-html" target="./build/logs/"/>` worked:

```
> bin/unit-tests-coverage
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

...........................................                       43 / 43 (100%)

44 sniff test files generated 120 unique error codes; 0 were fixable (0%)

Time: 7.18 seconds, Memory: 20.00 MB

OK (43 tests, 0 assertions)

Generating code coverage report in HTML format ... done
```

👏

GaryJones:
`./build/logs` and `./build/logs/` both worked for me.

jrfnl:
@rebeccahum @GaryJones Thank you both for testing. Seems that even PHPUnit has some cross-platform quirks.

I've added the `.` now to the config file.

Issue #358, opened 2019-01-16
Hooks logic fails for short array syntax


GaryJones:
Changing to short array syntax in `Tests/Hooks/AlwaysReturnInFilterUnitTests.inc` and `Tests/Hooks/PreGetPostsUnitTests.inc` result in those tests failing.

It suggests that some of the processing logic in each sniff is flawed for short array syntaxes.

Both long and short array syntax should be retained in the unit tests.

## Environment

Use `php -v` and `composer show` to get versions.

| Question                | Answer
| ------------------------| -------
| PHP version             | 7.2.0
| PHP_CodeSniffer version | 3.4.0
| VIPCS version           | `master`

## Tested Against `master` branch?

- [x] I have verified the issue still exists in the `master` branch of VIPCS.

GaryJones:
It's not as simple as adding a check for `T_OPEN_SHORT_ARRAY`.

Ideally, we want to be able to use [this WPCS method](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/e73a10e1c12ce07a12c3a8756efc54701458d23b/WordPress/Sniff.php#L2014) to accurately handle this.

See #359 for how this might be possible.

GaryJones:
Now that #377 is merged, we can use the WPCS method to fix/improve our sniff.

GaryJones:
WPCS 3.0 will likely that utility method, as it will depend on the utility methods in PHPCSUtils instead.

Good news is, after #474 is done, we can use these PHPCSUtils utility methods ourselves.

- [Long array syntax currently in the test file](https://github.com/Automattic/VIP-Coding-Standards/blob/478f56cc26f9eccf74f22a6ad994ba197105d2b1/WordPressVIPMinimum/Tests/Hooks/AlwaysReturnInFilterUnitTest.inc#L29-L64)
- [AlwaysReturnInFilterSniff](https://github.com/Automattic/VIP-Coding-Standards/blob/develop/WordPressVIPMinimum/Sniffs/Hooks/AlwaysReturnInFilterSniff.php)

jrfnl:
As #563 and #564 are now both merged, I believe this issue is fixed an can be closed.

PR #564, opened 2020-07-28
Hooks/PreGetPosts: add support for hook-ins using short arrays

jrfnl:
As VIPCS is currently using WPCS 2.x, we can use the WPCS `Sniff::find_array_open_close()` method to get the opener/closer for an array independently of the type of array (long/short).

Once VIPCS implements PHPCSUtils, this method call should be swopped out for the PHPCSUtils `Arrays::getOpenClose()` method.

Addresses #358 for the `PreGetPosts` sniff.

Includes unit tests.


GaryJones:
This one now has a merge conflict.

jrfnl:
Yup, I expected that one. Rebased & force pushed.

Issue #518, opened 2020-07-27
Review the WordPressVIPMinimum.Functions.RestrictedFunctions sniff


jrfnl:
Review the `WordPressVIPMinimum.Functions.RestrictedFunctions` sniff for the following in as far as relevant to that sniff:

- [ ] Code style independent sniffing / Correct handling of quirky code
    Typical things to add tests for and verify correct handling of:
    - [ ] Nested function/closure declarations
    - [ ] Nested class declarations
    - [ ] Comments in unexpected places
    - [ ] Variables being assigned to via `list` statements
    - [ ] Multiline text strings
    - [ ] Text strings provided via heredoc/nowdoc
    - [ ] Use of short open tags
    - [ ] Using PHP close tag as end of statement
- [ ] Code simplifications which can be made using PHPCSUtils
- [ ] Sniff stability improvements which can be made using PHPCSUtils
- [ ] Correct handling of modern PHP code
    Typical things to add tests for and verify correct handling of (where applicable):
    - [ ] PHP 5.0 Try/catch/finally (PHP 5.5) and exceptions
    - [ ] PHP 5.3 Namespaced code vs code in the global namespace
    - [ ] PHP 5.3 Use import statements, incl aliasing
    - [ ] PHP 5.3 Short ternaries
    - [ ] PHP 5.3 Closures, incl closure use
    - [ ] PHP 5.4 Short arrays
    - [ ] PHP 5.5 Class name resolution using `::class`
    - [ ] PHP 5.5 List in foreach
    - [ ] PHP 5.5/7.0 Generators using yield and yield from
    - [ ] PHP 5.6 Constant scalar expressions
    - [ ] PHP 5.6 Importing via `use function/const`
    - [ ] PHP 7.0 Null coalesce
    - [ ] PHP 7.0 Anonymous classes
    - [ ] PHP 7.0 Scalar and return type declarations
    - [ ] PHP 7.0 Group use statements
    - [ ] PHP 7.1 Short lists
    - [ ] PHP 7.1 Keyed lists
    - [ ] PHP 7.1 Multi-catch
    - [ ] PHP 7.1 Nullable types
    - [ ] PHP 7.3 List reference assignments
    - [ ] PHP 7.4 arrow functions
    - [ ] PHP 7.4 numeric literals with underscores
    - [ ] PHP 7.4 null coalesce equals
    - [ ] PHP 7.4 Typed properties
[...]

PR #566, opened 2020-07-28
RestrictedFunctions: fix false negative - functions in `config_settings` would never match

jrfnl:
The functions in the `config_settings` group were using a unicode `_`, not the ascii `_`, so these would never match the PHP native `opcache_*` functions they were intended to match.

As the same unicode character was used in the applicable unit tests and ruleset tests, the bug was never reported as a test failure.

Related to #518 (sniff review)


rebeccahum:
Nice!

PR #569, opened 2020-07-29
RestrictedVariables: don't report on "use" in `isset()`

### LINKED ISSUE #568: Check for RestrictedVariables wrong
e.g. WordPressVIPMinimum.Variables.RestrictedVariables.session___SESSION
will trigger for code like:

```
// dont cache it if we have an active PHP session
if ( isset( $_SESSION ) ) {
	$cache = false;
}
```

which doesn't make sense, as it's not even being used.
Afaik WP PHPCS had the same issue, but they fixed it already quite some time ago

jrfnl:
Disregard when the existence of a restricted variable is being checked. This uses the upstream WPCS `Sniff::is_in_isset_or_empty()` method.

This means that variables will not be reported as "used" when they are wrapped in a call to:
* `isset()`
* `empty()`
* `array_key_exists()`

This also means that the `if ( isset( $_SERVER['REMOTE_ADDR'] ) ) {` test on line 13 will no longer report a warning.

As `$_SERVER['REMOTE_ADDR']` was then no longer tested for and `$_SERVER['HTTP_USER_AGENT']` was being tested twice (line 14 and line 28), I've changed the occurrence on line 14 to use `$_SERVER['REMOTE_ADDR']` to make sure both are still tested.

Includes additional unit test for the case as reported by the op.

Fixes #568

### DIFF
--- a/WordPress-VIP-Go/ruleset-test.inc
+++ b/WordPress-VIP-Go/ruleset-test.inc
@@ -53,7 +53,7 @@ setcookie( 'cookie[three]', 'cookiethree' ); // Error + Message.
 $x = sanitize_key( $_COOKIE['bar'] ); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotValidated -- Error + Message.
 
 // WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__HTTP_USER_AGENT__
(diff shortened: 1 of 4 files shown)


jrfnl:
Note: this reminds me that I should open a review issue for the abstract sniff ;-)

PR #571, opened 2020-07-30
CS: remove yoda conditions from VIPCS codebase

jrfnl:
### PHPCS ruleset: forbid yoda conditions in VIPCS codebase

In contrast to WPCS, disallow the use of Yoda conditions.

### CS: change yoda conditions to non-yoda

Fixes made with the auto-fixer available in the Slevomat standard via the `SlevomatCodingStandard.ControlStructures.DisallowYodaComparison` sniff and visually reviewed.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -17,6 +17,7 @@
 		<exclude name="WordPress.Files.FileName"/>
 		<exclude name="WordPress.NamingConventions.ValidVariableName"/>
 		<exclude name="Generic.Arrays.DisallowShortArraySyntax"/>
+		<exclude name="WordPress.PHP.YodaConditions"/>
 	</rule>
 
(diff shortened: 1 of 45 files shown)


jrfnl:
@rebeccahum Please see the discussion Gary and me had about this on Slack for context.

rebeccahum:
d5v1hk8-00f17bce-6077-413a-8f24-b7909ec713c0

Since it seems like yoda is a preferential coding style, I'm fine with disabling the warning on the non-usage of it. However, is it necessary to enforce the usage of it when writing sniffs (I'm assuming for consistency)?

jrfnl:
> However, is it necessary to enforce the usage of it when writing sniffs (I'm assuming for consistency)?

@rebeccahum I presume you mean "enforce that yoda conditions aren't used" ?

Necessary: no, helpful for keeping the codebase consistent: yes.

PR #576, opened 2020-08-31
Docs: Update notes for releasing

GaryJones:
Documenting our release process, and updating the release template slightly.


jrfnl:
Just a quick question: as VA is now a dependency and more dependencies are expected to be added in the near future, should there be an action item to verify the version constraints of dependencies and update those if needed ?

PR #583, opened 2020-09-13
Composer: require the DealerDirect plugin

jrfnl:
As VIPCS requires two external standards, including the new `VariableAnalysis` dependency, let's require the DealerDirect plugin to make life easier on people who install via Composer.

Includes updating the Readme to mention the `VariableAnalysis` standard, as well as mention that the DealerDirect plugin is now a project requirement.

Note: I've widened the version constraints for the DealerDirect plugin to prevent conflicts with customer projects which already required the plugin, but potentially at a different version.
The version constraints now set covers all released versions which support external standards properly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,17 +16,14 @@
 	],
 	"require": {
 		"php": ">=5.4",
+		"dealerdirect/phpcodesniffer-composer-installer": "^0.4.1 || ^0.5 || ^0.6.2 || ^0.7",
 		"sirbrillig/phpcs-variable-analysis": "^2.8.3",
 		"squizlabs/php_codesniffer": "^3.5.5",
 		"wp-coding-standards/wpcs": "^2.3"
 	},
 	"require-dev": {
-		"dealerdirect/phpcodesniffer-composer-installer": "^0.7",
 		"phpcompatibility/php-compatibility": "^9",
 		"phpunit/phpunit": "^4 || ^5 || ^6 || ^7"
-	},
-	"suggest": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -18,16 +18,13 @@
 		"php": ">=5.4",
 		"sirbrillig/phpcs-variable-analysis": "^2.8.3",
 		"squizlabs/php_codesniffer": "^3.5.5",
-		"wp-coding-standards/wpcs": "^2.3"
+		"wp-coding-standards/wpcs": "^2.3",
+		"dealerdirect/phpcodesniffer-composer-installer": "^0.4.1 || ^0.5 || ^0.6.2 || ^0.7"
```
GaryJones, line 22: Let's [normalize](https://github.com/ergebnis/composer-normalize) this into alphabetical order please.

jrfnl: In that case, would it make sense to add that package to `require-dev` and add the `--dry-run` check to the Travis `lint` stage (with Composer script to ensure the same settings are used locally and in CI) ?

jrfnl: Ok, just checked - adding it to `require-dev` is not an option as the PHP requirements would conflict. Still, we could just install it temporarily for the CI dry-run only.

jrfnl: Either way, I've updated the commit to "normalize" the package order for now.

GaryJones: I'm not fussed about adding it to the dev dependencies, or having Travis fail a build because of it - it's rare to add new dependencies, so if we happen to catch it manually, great, but if not, not a biggie.

GaryJones (approved):

PR #584, opened 2020-09-14
Performance: more selective sniffing for efficiency

jrfnl:
The `register()` method tells PHPCS which tokens a sniff is listening for.

A number of sniffs were currently registering the [`Tokens::$functionNameTokens`](https://github.com/squizlabs/PHP_CodeSniffer/blob/9cce6859a595b3fa77fc823d9b0d451951dcf9d2/src/Util/Tokens.php#L570-L583) array.
In all cases, these sniffs were only looking for specific function calls, and in most cases, a "name" check is the first thing in the sniff code or else such a check was done at a later point in the code.

This makes using the `Tokens::$functionNameTokens` array highly inefficient, as the _only_ token within that array which could have the "name" the sniff is looking for is `T_STRING`. The other eleven (!) tokens registered would **never** match and would never trigger an error, but would be passed to the sniff even so.

This commit replaces the use of the `Tokens::$functionNameTokens` array in nearly all places in the code base with `T_STRING` for higher accuracy and more efficient sniffing.

**Note**: in a future PR, once PHPCSUtils is used, these sniffs should be adjusted to take the [PHP 8.0 namespaced identifier name tokens](https://wiki.php.net/rfc/namespaced_names_as_token) into account, but that is for later.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Compatibility/ZoninatorSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Compatibility/ZoninatorSniff.php
@@ -22,7 +22,7 @@ class ZoninatorSniff extends Sniff {
 	 * @return array(int)
 	 */
 	public function register() {
-		return Tokens::$functionNameTokens;
+		return [ T_STRING ];
 	}
 
 
--- a/WordPressVIPMinimum/Sniffs/Functions/CheckReturnValueSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/CheckReturnValueSniff.php
@@ -60,7 +60,7 @@ class CheckReturnValueSniff extends Sniff {
 	 * @return array(int)
 	 */
 	public function register() {
-		return Tokens::$functionNameTokens;
+		return [ T_STRING ];
 	}
 
 	/**
@@ -86,7 +86,7 @@ class CheckReturnValueSniff extends Sniff {
 	 */
 	private function isFunctionCall( $stackPtr ) {
 
-		if ( in_array( $this->tokens[ $stackPtr ]['code'], Tokens::$functionNameTokens, true ) === false ) {
+		if ( $this->tokens[ $stackPtr ]['code'] !== T_STRING ) {
 			return false;
 		}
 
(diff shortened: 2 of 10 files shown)


rebeccahum (approved):

GaryJones:
I like it!

Is there any metrics on what resources this changes saves? i.e. if WordPress or other large code base is scanned before and after, how much time / memory, on average is saved?

jrfnl:
> Is there any metrics on what resources this changes saves? i.e. if WordPress or other large code base is scanned before and after, how much time / memory, on average is saved?

I didn't run any metrics and whether it really makes a difference very much depends on the code used in a code base.

https://github.com/squizlabs/PHP_CodeSniffer/blob/9cce6859a595b3fa77fc823d9b0d451951dcf9d2/src/Util/Tokens.php#L570-L583

Of this list of Tokens, I suspect `T_STRING` will occur most often anyway, as it is used for all sorts of things, like function names (both in calls as well as declarations), class names, namespace names, constant names etc.

* `eval` shouldn't occur much, if at all, due to the security implications
* The `include|require[_once]` tokens will generally only be found in "old school" code which doesn't use autoloading or in code loading templates. Again, probably not that often.
* `isset()`, `unset()` and `empty()` will occur a lot in any code base which uses proper defensive coding, so there's a definite saving there.
* `self` and `static` will occur regularly in OO code, so again, there is a saving there.

Looking at the list now, I very much wonder why `parent` is not included, while `self` and `static` _are_, but that's a issue for upstream.

If you want some stats, I could do a run with Blackfire over, for instance, WP core with just and only this sniff, before and after this commit. Let me know.

GaryJones:
Metrics would be nice, but not essential, just to see what a change like this actually (approximately) gives in terms of something folks could understand.

jrfnl:
@GaryJones I'll do a run for the sake of it. May take a little while. I'll post results when I have them.

jrfnl:
Ok, so here goes... note: the full profiles via the links below are only available for 24 hours!

* [Run 1 on the `develop` commit from which this PR was branched off](https://blackfire.io/profiles/00373513-9867-4a22-b105-b5e5750e9363/graph).
    - Wall Time  2min 29s
    - Memory       61.4MB
    - 150296 calls to the `process` method of each of those sniffs
* [Run 2 on the commit in this PR](https://blackfire.io/profiles/c0a012a0-d8e5-403d-8aa7-9bcf13d08c2c/graph).
    - Wall Time   2min 4s
    - Memory       61.3MB
    - 133465 calls to the `process` method of each of those sniffs
* [Comparison between the runs](https://blackfire.io/profiles/compare/00373513-9867-4a22-b105-b5e5750e9363...c0a012a0-d8e5-403d-8aa7-9bcf13d08c2c/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=&callname=main()).
    image

Details:
* Command used and run against WP Core (src directory only):
   ```
   phpcs -ps ./src/ --extensions=php --report=summary,source --standard=wordpressvipminimum --sniffs=WordPressVIPMinimum.compatibility.zoninator,WordPressVIPMinimum.unctions.checkreturnvalue,WordPressVIPMinimum.hooks.alwaysreturninfilter,WordPressVIPMinimum.hooks.pregetposts,WordPressVIPMinimum.performance.cachevalueoverride,WordPressVIPMinimum.performance.fetchingremotedata,WordPressVIPMinimum.performance.taxonomymetainoptions,WordPressVIPMinimum.security.exitafterredirect,WordPressVIPMinimum.security.properescapingfunction,WordPressVIPMinimum.security.staticstrreplace
   ```
* Sample size: 3, i.e. the command was run three times for each profile.

Oh and I, just now, notice a typo in the command above (`unctions` not `functions`), so this profile is incomplete and the real performance gain will be even higher, but even if it were 16%, I'd say that's not bad going for one PR ;-)

GaryJones:
Great, thanks for that!

Presumably the listed sniffs are the ones that got amended in this PR?

jrfnl:
Exactly (aside from one not having been used due to the typo)

PR #589, opened 2020-10-19
WPQueryParams: sniff for `exclude` array key being set

### LINKED ISSUE #369: Flag 'exclude' args in get_posts()

Since `WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn` checks for `post__not_in`, we should also check for the `'exclude'` parameter being passed into `get_posts()` which in turn calls `post__not_in`.
```php
get_posts( [ 'exclude' => $post_ids ] ); // Warning.
```
```php
$exclude = [ 1, 2, 3 ];
```
## Additional context
Codex: https://developer.wordpress.org/reference/functions/get_posts/

jrfnl:
Note: I've split this change into two commits to allow for moving the second commit to the 3.0.0 release if so desired.

---

### WPQueryParams: sniff for `exclude` array key being set

This:
* Switches the `WordPressVIPMinimum.Performance.WPQueryParams` sniff over to use the WordPressCS `AbstractArrayAssignmentRestrictionsSniff` as the parent sniff.
* Adds sniffing for the `exclude` array key via the `AbstractArrayAssignmentRestrictionsSniff` sniff logic.

Includes unit test.

Fixes #369

### WPQueryParams: defer to the parent sniff

This changes over the existing keys being sniffed for by this sniff to use the logic from the new `AbstractArrayAssignmentRestrictionsSniff` parent instead of custom logic.

Note: this is a BC-break as the error codes for the existing checks change!
* `WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn` is now `WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_post__not_in`
* `WordPressVIPMinimum.Performance.WPQueryParams.SuppressFiltersTrue` is now `WordPressVIPMinimum.Performance.WPQueryParams.SuppressFilters_suppress_filters`

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Performance/WPQueryParamsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Performance/WPQueryParamsSniff.php
@@ -8,7 +8,7 @@
 
 namespace WordPressVIPMinimum\Sniffs\Performance;
 
-use WordPressVIPMinimum\Sniffs\Sniff;
+use WordPressCS\WordPress\AbstractArrayAssignmentRestrictionsSniff;
 use PHP_CodeSniffer\Util\Tokens;
 
 /**
@@ -16,7 +16,7 @@ use PHP_CodeSniffer\Util\Tokens;
  *
  *  @package VIPCS\WordPressVIPMinimum
  */
-class WPQueryParamsSniff extends Sniff {
+class WPQueryParamsSniff extends AbstractArrayAssignmentRestrictionsSniff {
 
 	/**
 	 * Returns an array of tokens this test wants to listen for.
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Performance/WPQueryParamsSniff.php
```diff
@@ -8,52 +8,86 @@
...
+				'keys'     => [
+					'suppress_filters',
+				],
+				'callback' => [ $this, 'value_is_true' ],
+			],
+			'PostNotIn' => [
+				'type'    => 'warning',
+				'message' => 'Using `post__not_in` should be done with caution, see https://wpvip.com/documentation/performance-improvements-by-removing-usage-of-post__not_in/ for more information.',
+				'keys'    => [
+					'post__not_in',
```
rebeccahum, line 51: Since `post__not_in` and `exclude` are the same, should we punt it into one group?

jrfnl: I considered that, but we'd need to have a careful think about the warning message in that case as using the existing message would be confusing.

rebeccahum: We can definitely revamp the warning message to something more ambiguous.

jrfnl: I'll have a think about it. Suggestions welcome.

rebeccahum: Maybe something like...
"Using exclusionary parameters should be done with caution, see https://wpvip.com/documentation/performance-improvements-by-removing-usage-of-post__not_in/ for more information."

We can even re-vamp the documentation a bit to account for it as well.

GaryJones:
As noted in #369, I'd prefer the second commit to be punted to 3.0 - thanks for thinking ahead and splitting out the code into the two commits!

jrfnl:
@GaryJones I've removed the second commit from the PR (kept locally with a "wait/3.0" annotation).

@rebeccahum  I've changed the group key for `exclude`  to `PostNotIn` so those two keys can be joined in one group with one message when we address part 2 of this issue.

Once this PR is merged, a new issues should be opened for 3.0 as a reminder that the rest of the sniff still needs to be switched over to the new abstract.

GaryJones (approved):
LGTM.

Issue #492, opened 2020-07-06
Remove BatCache related sniff


GaryJones:
## What problem would the enhancement address for VIP?

~Words matter, and the current understanding is that _whitelist_ has negative connotations for some people that can be easily avoided by using a different term.~

BatCache is not used on the current VIP platform, so this sniff is no longer needed.

## Describe the solution you'd like

~Change class, class properties and any other references to _whitelist_ for something else.~

Remove the sniff.

## Additional context

~Apart from the changelog, the only other references seem to be in relation to the BatcacheWhitelistedParams sniff and related test files.~

rebeccahum:
+1000000

GaryJones:
This _is_ important, but changing the class name will create a backwards compatibility break, so that means it needs to be done on a major release.

It's possible that all of the VIP clients will have moved over from WordPress.com VIP to VIP Go platform by the time the VIPCS is released, which means that this issue can change to removing this sniff completely, since it's only used in the WordPress-VIP-Minimum ruleset.

GaryJones:
> It's possible that all of the VIP clients will have moved over from WordPress.com VIP to VIP Go platform by the time the VIPCS is released, which means that this issue can change to removing this sniff completely, since it's only used in the WordPress-VIP-Minimum ruleset.

This is now correct, so the BatCache-related sniff can be removed completely now.

GaryJones:
Closing in favour of #613.

Issue #517, opened 2020-07-27
Review the WordPressVIPMinimum.Functions.DynamicCalls sniff


jrfnl:
Review the `WordPressVIPMinimum.Functions.DynamicCalls` sniff for the following in as far as relevant to that sniff:

- [ ] Code style independent sniffing / Correct handling of quirky code
    Typical things to add tests for and verify correct handling of:
    - [ ] Nested function/closure declarations
    - [ ] Nested class declarations
    - [ ] Comments in unexpected places
    - [ ] Variables being assigned to via `list` statements
    - [ ] Multiline text strings
    - [ ] Text strings provided via heredoc/nowdoc
    - [ ] Use of short open tags
    - [ ] Using PHP close tag as end of statement
- [ ] Code simplifications which can be made using PHPCSUtils
- [ ] Sniff stability improvements which can be made using PHPCSUtils
- [ ] Correct handling of modern PHP code
    Typical things to add tests for and verify correct handling of (where applicable):
    - [ ] PHP 5.0 Try/catch/finally (PHP 5.5) and exceptions
    - [ ] PHP 5.3 Namespaced code vs code in the global namespace
    - [ ] PHP 5.3 Use import statements, incl aliasing
    - [ ] PHP 5.3 Short ternaries
    - [ ] PHP 5.3 Closures, incl closure use
    - [ ] PHP 5.4 Short arrays
    - [ ] PHP 5.5 Class name resolution using `::class`
    - [ ] PHP 5.5 List in foreach
    - [ ] PHP 5.5/7.0 Generators using yield and yield from
    - [ ] PHP 5.6 Constant scalar expressions
    - [ ] PHP 5.6 Importing via `use function/const`
    - [ ] PHP 7.0 Null coalesce
    - [ ] PHP 7.0 Anonymous classes
    - [ ] PHP 7.0 Scalar and return type declarations
    - [ ] PHP 7.0 Group use statements
    - [ ] PHP 7.1 Short lists
    - [ ] PHP 7.1 Keyed lists
    - [ ] PHP 7.1 Multi-catch
    - [ ] PHP 7.1 Nullable types
    - [ ] PHP 7.3 List reference assignments
    - [ ] PHP 7.4 arrow functions
    - [ ] PHP 7.4 numeric literals with underscores
    - [ ] PHP 7.4 null coalesce equals
    - [ ] PHP 7.4 Typed properties
[...]

PR #592, opened 2020-10-23
Functions/DynamicCalls: various bug fixes and improvements

### LINKED ISSUE #590: Undefined offset: 42509 in /opt/composer/vendor/automattic/vipwpcs/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php on line 218
## Bug Description

I get the following notice when using the PHPCS VIP Sniff. The issue seems to be file specific though:
Undefined offset: 42509 in /opt/composer/vendor/automattic/vipwpcs/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php on line 218

Trace:

```
File: \opt\composer\vendor\squizlabs\php_codesniffer\src\Runner.php on line 606
Trace:
#0 /opt/composer/vendor/automattic/vipwpcs/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php (218): PHP_CodeSniffer\Runner->handleErrors( $code = 8888888, $message = Undefined offset: 42509, $file = /opt/composer/vendor/automattic/vipwpcs/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php, $line = 218, Array( [i] => 1 ) )
#1 /opt/composer/vendor/automattic/vipwpcs/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php (86): WordPressVIPMinimum\Sniffs\Functions\DynamicCallsSniff->find_dynamic_calls()
#2 /opt/composer/vendor/wp-coding-standards/wpcs/WordPress/Sniff.php (910): WordPressVIPMinimum\Sniffs\Functions\DynamicCallsSniff->process_token( $stackPtr = 42508 )
#3 /opt/composer/vendor/squizlabs/php_codesniffer/src/Files/File.php (496): WordPressCS\WordPress\Sniff->process( $phpcsFile = PHP_CodeSniffer\Files\LocalFile Object, $stackPtr = 42508 )
#4 /opt/composer/vendor/squizlabs/php_codesniffer/src/Files/LocalFile.php (91): PHP_CodeSniffer\Files\File->process()
#5 /opt/composer/vendor/squizlabs/php_codesniffer/src/Fixer.php (174): PHP_CodeSniffer\Files\LocalFile->process()
#6 /opt/composer/vendor/squizlabs/php_codesniffer/src/Reports/Cbf.php (49): PHP_CodeSniffer\Fixer->fixFile()
#7 /opt/composer/vendor/squizlabs/php_codesniffer/src/Reporter.php (285): PHP_CodeSniffer\Reports\Cbf->generateFileReport( $report =
Array(
 [filename] => /my-path/my-file.php
 [errors] => 843
 [warnings] => 205
 [fixable] => 816
 [messages] => Array(
   [1] => Array(
     [1] => Array(
       [0] => Array(... TRUNCATED
, $phpcsFile = PHP_CodeSniffer\Files\LocalFile ObjectPHP_CodeSniffer\Files\LocalFile ObjectPHP_CodeSniffer\Files\LocalFile ObjectPHP_CodeSniffer\Files\LocalFile ObjectPHP_CodeSniffer\Files\LocalFile Object, $showSources = false, $width = 0 )
#8 /opt/composer/vendor/squizlabs/php_codesniffer/src/Runner.php (656): PHP_CodeSniffer\Reporter->cacheFileReport( $phpcsFile = PHP_CodeSniffer\Files\LocalFile Object, PHP_CodeSniffer\Config Object )
#9 /opt/composer/vendor/squizlabs/php_codesniffer/src/Runner.php (434): PHP_CodeSniffer\Runner->processFile( $file = PHP_CodeSniffer\Files\LocalFile Object )
#10 /opt/composer/vendor/squizlabs/php_codesniffer/src/Runner.php (200): PHP_CodeSniffer\Runner->run()
#11 /opt/composer/vendor/squizlabs/php_codesniffer/bin/phpcbf (18): PHP_CodeSniffer\Runner->runPHPCBF()
```

## Minimal Code Snippet

Can you see if you can check it with the above trace? As this was a development file and we don't have that version anymore.

## Environment

| PHP version             | PHP 7.4.11
| PHP_CodeSniffer version | 3.5.6
| VIPCS version           | 2.2.0

jrfnl:
This PR is the result of a review of the `WordPressVIPMinimum.Functions.DynamicCalls` sniff.

It will be easiest to understand the different changes and their impact by reviewing this PR on the individual commits.

---

### Functions/DynamicCalls: annotate the unit tests

... to show which ones should fail and which should pass.

### Functions/DynamicCalls: remove redundant conditions [1]

This sniff only listens to `T_VARIABLE` tokens, so checking that what was received is a `T_VARIABLE` is redundant.

### Functions/DynamicCalls: remove redundant conditions [2]

The condition above checks if a token is `T_EQUAL` and bows out if it is not.
The only code matching on `T_EQUAL` is the `=` operator, which will always have a `length` of `1`.

### Functions/DynamicCalls: improve code readability

* Cutting code lines off at 50 chars is maybe taking it a little too far, especially as it makes assignments hard to read.
* Use proper tags in docblocks.
* Improve (fix) documentation (and move it to the right place).

### Functions/DynamicCalls: minor simplification

Join two conditions which both return anyway.

### Functions/DynamicCalls: bug fix - ignore comments [1]

Skip over both whitespace, as well as comments. This reduces false negatives.

Includes unit test.

### Functions/DynamicCalls: bug fix - don't blindly use the next text string

A variable value may be build up of multiple tokens.

As it was, the sniff would look for the first text string token after the equal sign within the variable assignment statement, but this disregards that:
1. The text string token found may not be the only token in the statement.
2. A statement can end on a PHP close tag (possibly a bug in PHPCS itself, but that's another matter), which would lead the sniff to look at the next statement for text strings.

Fixed now.

Includes unit tests, the first four of which resulted in false positives previously.

### Functions/DynamicCalls: bug fix - allow for double quotes

Text strings can use both single quotes as well as double quotes.

When the text string contains an interpolated variable, it will be tokenized as `T_DOUBLE_QUOTED_STRING`, but when it is a plain text string, a double quoted text string will be tokenized as `T_CONSTANT_ENCAPSED_STRING`, same as single quoted text string.

The sniff did not take this into account, leading to false negatives.

The sniff also would strip quotes from within a text - `'my\'text'` - . This did not cause a problem for this sniff as function names cannot have back slashes in them, but it was still wrong.

Fixed now by using the WPCS `strip_quotes()` method.

Includes unit test which would fail previously.

### Functions/DynamicCalls: bug fix - fix memory + performance issue

The sniff maintains a cache of all the variables it has seen and their assigned value.

When a variable is encountered, it would:
* Check if it was an (plain text) assignment and if so, register the variable name + value to the cache.
* Next, call the `find_dynamic_calls()` method, which first checks if any variables have been registered to the cache before doing anything.
* And then checks for dynamic function calls and if one is found, checks if the variable used is one registered in the cache with a value we are looking for.

This is highly inefficient as text string variable assignments are common and, as it was, _every single one_ would be added to the cache.

With a large code base, that means that the cache could grow pretty large.

It also means that the logic to determine if something is a dynamic function call would be executed even when there would be no text strings registered in the cache which could match any of the ones we're looking for.

By changing the order of the logic, the memory leak and performance inefficiency is removed.

With the updated logic, the sniff will:
* Check if it was an (plain text) assignment **and if the text string matches one we're looking for** and if so, register the variable name + value to the cache.
* Next, call the `find_dynamic_calls()` method, which first checks if any variables have been registered to the cache before doing anything.
* And then checks for dynamic function calls.

This means that if none of the previous assignments encountered matches any of the target text strings (~ 99% of the time), this sniff will bow out at step 2 before executing the logic to check if a variable assignment is a dynamic function call.

### Functions/DynamicCalls: rename private property

Rename the `private` `$blacklisted_functions` property to `$function_names` to get rid of the use of a non-inclusive term.

Loosely related to #492

### Functions/DynamicCalls: bug fix - ignore comments [2]

Skip over both whitespace, as well as comments and take live coding into account. This reduces false negatives, as well as fixing issue #590.

Includes unit tests.

Fixes #590

### Functions/DynamicCalls: error message tweak

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
@@ -31,15 +31,15 @@ class DynamicCallsSniff extends Sniff {
 	 * @var array
 	 */
 	private $blacklisted_functions = [
-		'assert',
-		'compact',
-		'extract',
-		'func_get_args',
-		'func_get_arg',
-		'func_num_args',
-		'get_defined_vars',
-		'mb_parse_str',
-		'parse_str',
+		'assert'           => true,
+		'compact'          => true,
+		'extract'          => true,
+		'func_get_args'    => true,
+		'func_get_arg'     => true,
+		'func_num_args'    => true,
+		'get_defined_vars' => true,
+		'mb_parse_str'     => true,
+		'parse_str'        => true,
 	];
 
 	/**
@@ -138,11 +138,17 @@ class DynamicCallsSniff extends Sniff {
 		/*
 		 * If we reached the end of the loop and the $value_ptr was set, we know for sure
 		 * this was a plain text string variable assignment.
-		 *
-		 * Register its name and value in the internal array for later usage.
 		 */
 		$current_var_value = $this->strip_quotes( $this->tokens[ $value_ptr ]['content'] );
 
+		if ( isset( $this->blacklisted_functions[ $current_var_value ] ) === false ) {
+			// Text string is not one of the ones we're looking for.
+			return;
+		}
+
+		/*
+		 * Register the variable name and value in the internal array for later usage.
+		 */
 		$this->variables_arr[ $current_var_name ] = $current_var_value;
 	}
 
@@ -183,19 +189,6 @@ class DynamicCallsSniff extends Sniff {
 
 		$t_item_key = $this->stackPtr + $i;
 
-		/*
-		 * We have a variable match, but make sure it contains name of a function which is on our blacklist.
-		 */
-
-		if ( ! in_array(
-			$this->variables_arr[ $this->tokens[ $this->stackPtr ]['content'] ],
-			$this->blacklisted_functions,
-			true
-		) ) {
-			return;
-		}
-
-		// We do, so report.
 		$message = 'Dynamic calling is not recommended in the case of %s.';
 		$data    = [ $this->variables_arr[ $this->tokens[ $this->stackPtr ]['content'] ] ];
 		$this->phpcsFile->addError( $message, $t_item_key, 'DynamicCalls', $data );
--- a/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
@@ -7,6 +7,7 @@
 
 namespace WordPressVIPMinimum\Sniffs\Functions;
 
+use PHP_CodeSniffer\Util\Tokens;
 use WordPressVIPMinimum\Sniffs\Sniff;
 
 /**
@@ -98,7 +99,7 @@ class DynamicCallsSniff extends Sniff {
 		 * and checking if the first one is T_EQUAL.
 		 */
 		$t_item_key = $this->phpcsFile->findNext(
-			[ T_WHITESPACE ],
+			Tokens::$emptyTokens,
 			$this->stackPtr + 1,
 			null,
 			true,
--- a/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
@@ -94,17 +94,6 @@ class DynamicCallsSniff extends Sniff {
 	 * @return void
 	 */
 	private function collect_variables() {
-		/*
-		 * Make sure we are working with a variable,
-		 * get its value if so.
-		 */
-
-		if (
-			$this->tokens[ $this->stackPtr ]['type'] !==
-				'T_VARIABLE'
-		) {
-			return;
-		}
 
 		$current_var_name = $this->tokens[ $this->stackPtr ]['content'];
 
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
```diff
@@ -120,10 +120,6 @@ private function collect_variables() {
 			return;
 		}
 
-		if ( $this->tokens[ $t_item_key ]['length'] !== 1 ) {
```
GaryJones, line 134: Just curious - what tokens are `==` and `===`?

GaryJones (approved):
Love the performance fix amongst the general fixes!

WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
jrfnl: * `==` => `T_IS_EQUAL`
* `===` => `T_IS_IDENTICAL`

Ref: https://www.php.net/manual/en/tokens.php

jrfnl:
Please also see my general review comment about this sniff which I've left in the review ticket: https://github.com/Automattic/VIP-Coding-Standards/issues/517#issuecomment-715566334

WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
```diff
@@ -7,47 +7,47 @@
...
-		'assert',
-		'compact',
-		'extract',
-		'func_get_args',
-		'func_get_arg',
-		'func_num_args',
-		'get_defined_vars',
-		'mb_parse_str',
-		'parse_str',
+	private $function_names = [
```
rebeccahum, line 33: Thank you for re-naming this!

PR #595, opened 2020-11-01
Security/Underscorejs: bug fixes and enhancements

### LINKED ISSUE #345: UnderscoreJS output escaping improvement?
The Underscorejs output escaping sniff checks for `<%=`, but it's possible that `<%= _.escape(...)` would also sufficiently escape the output.

I don't know UnderscoreJS, so this needs looking into, but it may help remove some false positives.

### LINKED ISSUE #546: Review the WordPressVIPMinimum.Security.Underscorejs sniff
Review the `WordPressVIPMinimum.Security.Underscorejs` sniff for the following in as far as relevant to that sniff:

    - [x] Multiline text strings
    - [x] Text strings provided via heredoc/nowdoc
    - [ ] Use of short open tags
    - [ ] PHP 5.3 Short ternaries
    - [ ] PHP 5.4 Short arrays
    - [ ] PHP 5.5 List in foreach
    - [ ] PHP 7.0 Null coalesce
    - [ ] PHP 7.1 Short lists
    - [ ] PHP 7.1 Keyed lists
    - [ ] PHP 7.1 Multi-catch
    - [ ] PHP 7.1 Nullable types
    - [ ] PHP 7.4 arrow functions

Other:
- [ ] Review PHPDoc comments

- [ ] PHP 8.0 Union types
- [ ] PHP 8.0 Named arguments

jrfnl:
Result of a review of the `WordPressVIPMinimum.Security.Underscorejs` sniff.

Not claiming completeness as JS, like PHP, can be written in lots of different ways - think ternaries, method chaining etc - and my JS is a bit rusty ;-).
Consider this a first round of improvements.

Input regarding underscores syntaxes based on:
* https://www.topjavatutorial.com/underscore-js/underscore-js-escape-unescape-functions/
* https://underscorejs.org/
* https://devdocs.io/underscore/index
* https://stackoverflow.com/questions/4778881/how-to-use-underscore-js-as-a-template-engine

Code samples largely based on samples found via:
* https://wpdirectory.net/search/01EN00V6N7CVNF8SXDW6EA91NV
* https://wpdirectory.net/search/01EN00VG3ERYZ2BZZ5VFFZ62N4
* https://wpdirectory.net/search/01EN375V8PHQYTZB9T39VB9X6R
* https://wpdirectory.net/search/01EN39JTD2A9W51AGPVZJ6ANHB

Fixes #546

---

### Security/Underscorejs: sniff all text string tokens

So far, only single quoted text strings, heredocs and inline HTML were sniffed.
However, there are two more text string token types: nowdoc and double quoted strings.

This adds these tokens to the `register()` method to be sniffed.

This commit also ensures that there is at least one unit test in place for each of these token types and that the tests use a variety of whitespace.

### Security/Underscorejs: throw an error for each violation

So far, the sniff would throw one error per text string, independently of how often the unescaped output notation would occur in the text string.

While for simple text strings, this is not much of an issue, for long and complex text strings, it may be more difficult to spot all the unescaped output notations in the text string.

This changes the sniff to:
* Throw a warning for each occurrence of the unescaped output notation in a text string.
* Include a snippet from the text string with the details of the unescaped output notation.
    Old: `Found Underscore.js unescaped output notation: "<%=".`
    New: `Found Underscore.js unescaped output notation: "<%= post_url %>".`

To allow the sniff to also match on a text string being concatenated together, the regex has to also match when `<%=` occurs at the end of the text string and will only display `<%=` in that case.
To this end, we also need to remove the text string quotes (if they exist) from the token content.

### Security/Underscorejs: add JS test case file

Add a basic JS test case file.

Tests for `interpolate` will be added in a follow-up commit.

### Security/Underscorejs: improve check for `interpolate` in JS files

The `interpolate` property in JS can be set using various syntaxes. While the current check will in no way catch them all, the changes now made should improve the accuracy of detecting a change in the delimiter via `interpolate` in JS files.

Notes:
* Add checking for `T_STRING` to detect `_.templateSettings.interpolate =` syntax. This was previously not detected (false negative).
* Add some defensive coding to prevent false positives when the `interpolate` keyword is used in another context than underscorejs.
* Limit the check for `T_STRING` and `T_PROPERTY` to Javascript files only to prevent false positives when scanning PHP files.
* Limit the check for `interpolate` in text strings to PHP only. This prevents false positives when the keyword is used in a text string in JS.

Includes unit tests covering all the above.

### Security/Underscorejs: improve check for `interpolate` in PHP files

Match the `interpolate` property in PHP files with the similar precision as in JS files.

Includes unit tests.

### Security/Underscorejs: bug fix - don't error when variable is escaped

Prevent triggering a warning when the variable being printed is escaped using `_.escape()`.

Includes unit tests.

Fixes #345

### Security/Underscorejs: ignore Gruntfile.js files

These are configuration files and not part of the production code.

This check does not verify whether this file is in the project root as we don't know what the project root is. It will plainly ignore any file called `Gruntfile.js` in a case-insensitive manner.

Includes unit tests.

### Security/Underscorejs: enhancement - check for print execution statements

Add a new check for the JS native `print` command and the `_p+=` variation, when used without being combined with `_.escape()`.

Includes unit tests.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Security/UnderscorejsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/UnderscorejsSniff.php
@@ -41,6 +41,7 @@ class UnderscorejsSniff extends Sniff {
 	public function register() {
 		$targets   = Tokens::$textStringTokens;
 		$targets[] = T_PROPERTY;
+		$targets[] = T_STRING;
 
 		return $targets;
 	}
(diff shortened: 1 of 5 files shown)


GaryJones (approved):
Love it!

PR #607, opened 2021-01-08
WordPress.Security.NonceVerification.NoNonceVerification: Remove outdated reference from WordPress-VIP-Go ruleset and let inheritance from parent

### LINKED ISSUE #605: Remove reference WordPress.Security.NonceVerification.NoNonceVerification
Per the changelog of WPCS 2.0.0-RC1:

> The WordPress.Security.NonceVerification sniff used the same error code for both an error as well as a warning. The old error code NoNonceVerification is no longer used. The error now uses the Missing error code, while the warning now uses the Recommended error code. If you are referencing the old error code in a ruleset XML file or in inline annotations, please update these to use the new codes instead.

I think we can remove the reference from the WordPress-VIP-Go ruleset: https://github.com/Automattic/VIP-Coding-Standards/blob/60ad148168258afccd05ff0429d53a7e810270a5/WordPress-VIP-Go/ruleset.xml#L132

The sniff is still inherited from the parent WordPressVIPMinimum (albeit, different error codes now):

https://github.com/Automattic/VIP-Coding-Standards/blob/60ad148168258afccd05ff0429d53a7e810270a5/WordPressVIPMinimum/ruleset.xml#L21

rebeccahum:
Fixes #605.

### DIFF
--- a/WordPress-VIP-Go/ruleset-test.inc
+++ b/WordPress-VIP-Go/ruleset-test.inc
@@ -67,7 +67,7 @@ $external_resource = file_get_contents( $test ); // Warning + Message.
 $file_content = file_get_contents( 'my-file.svg' ); // Ok.
 wpcom_vip_file_get_contents( $bar ); // Ok.
 
-// WordPress.Security.NonceVerification.NoNonceVerification
+// WordPress.Security.NonceVerification (inherited from parent)
 function bar_foo() {
 	if ( ! isset( $_POST['test'] ) ) { // Error.
 		return;
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -129,11 +129,6 @@
 
 
 	<!-- Warnings and other things -->
-	<rule ref="WordPress.Security.NonceVerification.NoNonceVerification">
-		<!-- Needs a manual check -->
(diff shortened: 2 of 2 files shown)


GaryJones (approved):
Approved - in this case, NoNonceVerification hasn't existed in WPCS since 2.0.0, and we've required WPCS 2.1.0 as a dev dependency since VIPCS 2.0, so all good.

jrfnl:
Reviewed after the fact.

Notes:

This commit constitutes a functional change, though in reality that change took effect once WPCS 2.0 was required via Composer.

Previously there was only one error code in WPCS from the `NonceVerification` sniff - `NoNonceVerification` -. This same error code would throw a `warning` or an `error` depending on whether or not a superglobal was referenced. Both would have severity `5`.

The ruleset change in VIP Go effectively, turned **all** notices from the `NonceVerification` sniff into `warning`s with severity 10.

Once WPCS 2.0 was required, the ruleset change in VIP Go no longer had any effect and the sniff would now throw the `Missing` error and the `Recommended` warning at severity `5`.

This update formalizes the current status-quo by removing the reference from the VIP Go ruleset.
The "old" situation of turning violations from the `NonceVerification` sniff into `warning`s with severity `10` is effectively abandoned by this change.

PR #610, opened 2021-01-08
WordPress-VIP-Go: Replace WordPress.WP.GlobalVariablesOverride.OverrideProhibited with WordPress.WP.GlobalVariablesOverride.Prohibited

### LINKED ISSUE #609: Update WordPress.WP.GlobalVariablesOverride.OverrideProhibited to WordPress.WP.GlobalVariablesOverride.Prohibited in WordPress-VIP-Go
As of WPCS 2.0.0-RC1:

> The WordPress.WP.GlobalVariablesOverride.OverrideProhibited error code has been replaced by the WordPress.WP.GlobalVariablesOverride.Prohibited error code. If you are referencing the old error code in a ruleset XML file or in inline annotations, you may need to update it.

We'll need to update it in `WordPress-VIP-Go`: https://github.com/Automattic/VIP-Coding-Standards/blob/60ad148168258afccd05ff0429d53a7e810270a5/WordPress-VIP-Go/ruleset.xml#L226

rebeccahum:
Fixes #609.

### DIFF
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -223,7 +223,7 @@
 		<!-- We trust that translations are safe on VIP Go -->
 		<severity>1</severity>
 	</rule>
-	<rule ref="WordPress.WP.GlobalVariablesOverride.OverrideProhibited">
+	<rule ref="WordPress.WP.GlobalVariablesOverride.Prohibited">
 		<!-- This is often a false positive. Still nice to flag for a check -->
 		<severity>3</severity>
 	</rule>


GaryJones:
Since it's been so long since this change happened, and we didn't notice, do we need an addition to a ruleset test to cover this as well?

rebeccahum:
It does look like we already do in the parent ruleset:

https://github.com/Automattic/VIP-Coding-Standards/blob/6d3f5267bee6f0e7548aacc3881b2ac0bcd981e9/WordPressVIPMinimum/ruleset-test.inc#L72-L74

I'm not sure what difference it would make for violation name changes that occur if it still triggers an expected error regardless.

GaryJones (approved):

jrfnl:
Reviewed after the fact.

To be quite fair, I'd recommend for this severity change to be removed completely. In reality it hasn't been active since the update to WPCS 2.0.0+ in April 2019 in commit 53505ffd7fcbd23beb8e91dbf4762259ee371a29 and AFAICS nobody has complained about it.

The reason why you probably haven't received many complaints is that the sniff has had quite a lot of improvements over the years which reduced false positives.

The only false positives which are common with this sniff is when using "views" which are themselves in stand-alone files (global namespace), but are in reality only used via an `include` from within a function. The variables declared in those views will, in practice, be local to the function doing the `include`, but that is something the sniff cannot analyse.
To exclude those, a property can be enabled. See: https://github.com/WordPress/WordPress-Coding-Standards/wiki/Customizable-sniff-properties#global-wp-variable-override-check-in-the-global-namespace

More than anything, the sniff currently has a lot of false negatives as a there are a lot of global variables in WP which the sniff is not yet looking for.

PR #611, opened 2021-01-08
Removing cruft since it is already a warning in WordPress.WP.AlternativeFunctions.file_system_read

rebeccahum:
Removing the cruft of declaring it as a warning since it is already one in the parent sniff:

https://github.com/WordPress/WordPress-Coding-Standards/blob/c52e511356c65c7ee97587dcb84e98b2f90e9633/WordPress/Sniffs/WP/AlternativeFunctionsSniff.php#L129-L143


GaryJones (approved):
Doesn't get amended in the parent WordPressVIPMinimum ruleset, so LGTM.

PR #617, opened 2021-01-11
Remove warning type since it inherits from parent as "warning" 

rebeccahum:
More cruft, similar to #611


GaryJones (approved):
It is indeed added as a [Warning in the sniff](https://github.com/Automattic/VIP-Coding-Standards/blob/6d3f5267bee6f0e7548aacc3881b2ac0bcd981e9/WordPressVIPMinimum/Sniffs/Performance/FetchingRemoteDataSniff.php#L49) and not overridden in the WordPressVIPMinimum ruleset.

Issue #613, opened 2021-01-11
Remove BatcacheWhitelistedParamsSniff


rebeccahum:
Same reason as #612 — no longer needed since WordPress.com sites are migrated.

jrfnl:
Same as for #612:

> Action plan suggestion:
> 1. Deprecate the sniff in the next release (2.x) by:
>     1. Adding a deprecation warning to the sniff which will inform people to remove the sniff reference from their own custom ruleset.
>     2. Silencing all messages from the sniff from within the `VIPMinimum` ruleset, effectively removing the sniff already for people using the standard rulesets.
> 2. Remove the actual sniff in the next major, i.e. 3.0.0.
> 
> Loosely related to #600

PR #618, opened 2021-01-11
Remove Batcache references in messaging

rebeccahum:
VIP Go does not utilize Batcache and we overwrite the messaging in the ruleset anyways, so I think we can clean it up a bit and remove the Batcache references.

#613 is related as well.


GaryJones (approved):
LGTM.

To summarise the changes here, as I see it:

 - Original sniffs were written for WPCOM, and so mentioned Batcache. 
 - The error messages were overridden in WordPress-VIP-Go ruleset to make them more generic (since Batcache isn't used there).
  - Since we no longer have any WPCOM, then the original sniff messages can be updated to be more generalised, and the ruleset overrides can be removed.

jrfnl:
Reviewed after the fact.

Notes:
* The change to the `WordPressVIPMinimum.Functions.RestrictedFunctions` `cookies` group is good, though there are some comments in the ruleset test case files which will still need to be updated.
    **Note**: this change will need a changelog entry as a `warning` has changed to `error` for the WordPressVIPMinimum standard.
* While not used in the VIP-Go standard, the message text for the `WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___SERVER__REMOTE_ADDR__` error code has now also been changed.
    That is not to say that the change is incorrect, but it is a change the PR makes no mention of.

I will send in a PR shortly to fix the comments in the ruleset test case files for both changes.

PR #620, opened 2021-01-18
Silence VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable

### LINKED ISSUE #498: Consider silencing the UnusedVariable violation
From #449:

> We could also take this opportunity is taken to silence the UnusedVariable violation. It's severity is already dropped down to 1 for `WordPress-VIP-Go`, which is the same as what the bot runs at (though this may change in the future). From a VIP perspective, unused variables are not important.

It would reduce the number of violation comments. Although it's not ideal to have values being to variables that are then unused (waste of memory), it's not a severe enough impact to typically worry about.

If clients were keen on this, they could run the VariableAnalyis sniff (or static analysis tools or the inspector plugin for PHPStorm), to find these.

Discuss and consider silencing the UnusedVariable violation.

...

...

rebeccahum:
Fixes #498.

### DIFF
--- a/WordPressVIPMinimum/ruleset.xml
+++ b/WordPressVIPMinimum/ruleset.xml
@@ -144,7 +144,9 @@
 	</rule>
 
 	<!-- Include VariableAnalysis checks -->
-	<rule ref="VariableAnalysis"/>
+	<rule ref="VariableAnalysis">
+		<exclude name="VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable"/>
+	</rule>
 
 	<!-- Prevent deprecation notice when the sniff is not explicitly included. -->
 	<rule ref="WordPressVIPMinimum.Variables.VariableAnalysis.DeprecatedSniff">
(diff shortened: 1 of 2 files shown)


WordPressVIPMinimum/ruleset.xml
```diff
@@ -144,7 +144,9 @@
 	</rule>
 
 	<!-- Include VariableAnalysis checks -->
-	<rule ref="VariableAnalysis"/>
+	<rule ref="VariableAnalysis">
+		<exclude name="VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable"/>
```
GaryJones, line 148: Instead of putting an exclude here, I would suggest just change the `1` in the other file to `0`.

See https://github.com/WordPress/WordPress-Coding-Standards/blob/c52e511356c65c7ee97587dcb84e98b2f90e9633/WordPress-Core/ruleset.xml#L70-L78 which is an example of using severity = 0, rather than exclude - See https://github.com/WordPress/WordPress-Coding-Standards/commit/7e5401cdc9e5d907b4f68ce9c80e754e8e200256 for the rationale.

GaryJones (changes requested):

GaryJones (approved):

PR #621, opened 2021-01-18
Add RestrictedCacheGroupSniff to restrict wp-memcached global cache groups

### LINKED ISSUE #260: Global memcached groups names
## Bug Description
We should not allow use of global memcached cache group names, as they are being used on a global scale and things go wrong (such as user login) when they are overwritten: https://github.com/Automattic/wp-memcached

## Minimal Code Snippet

```php
wp_cache_set( 'blahblah', 'blah', 'users', 1 * MINUTE_IN_SECONDS );
```

rebeccahum:
Fixes #260


WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
```diff
@@ -0,0 +1,100 @@
...
+		'wp_cache_set' => true,
+		'wp_cache_add' => true,
+	];
+
+	/**
+	 * List of cache group names already in use by wp-memcached.
+	 *
+	 * @var array
+	 */
+	private $wp_memcached_groups = [
```
GaryJones, line 46: Can these be put into alphabetical order?

WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
```diff
@@ -0,0 +1,100 @@
...
+	 * @param int    $stackPtr        The position of the current token in the stack.
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 * @return int|void Integer stack pointer to skip forward or void to continue
+	 *                  normal file processing.
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		if ( count( $parameters ) > 2 && isset( $this->wp_memcached_groups[ trim( $parameters[3]['raw'], '"\'' ) ] ) ) {
+			$this->phpcsFile->addError( 'Please do not use cache group %s, as it is already in use by wp-memcached: https://docs.wpvip.com/technical-references/caching/object-cache/.', $stackPtr, 'wp_memcached', $parameters[3]['raw'] );
```
GaryJones, line 97: Should this violation code follow the typical format, of starting with an uppercase character?

WordPressVIPMinimum/Tests/Compatibility/RestrictedCacheGroupUnitTest.inc
```diff
@@ -0,0 +1,15 @@
...
+wp_cache_set( 'cache_key', $data, 'users', 10 * MINUTE_IN_SECONDS );
+wp_cache_set( 'cache_key', $data, "userlogins" );
+wp_cache_add( $cache_key, 'data', 'posts' );
+wp_cache_add( $cache_key,   $data,   "category_relationships", 600000 );
+
+// OK
+$key = 'cache_key';
+wp_cache_set( $key, $data, 'users_stuff' );
+wp_cache_add( $key, 'data', "posts_" );
+wp_cache_set( $key, $data, 'custom_taxonomy_relationships' );
```
GaryJones, line 15: How about some more unit tests with extra line breaks, comments and so on between args, to show that it is code-style agnostic?

GaryJones (changes requested):

WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
```diff
@@ -0,0 +1,99 @@
...
+	 * @param array  $group_name      The name of the group which was matched.
+	 * @param string $matched_content The token content (function name) which was matched.
+	 * @param array  $parameters      Array with information about the parameters.
+	 * @return int|void Integer stack pointer to skip forward or void to continue
+	 *                  normal file processing.
+	 */
+	public function process_parameters( $stackPtr, $group_name, $matched_content, $parameters ) {
+		if ( count( $parameters ) > 2 && isset( $this->wp_memcached_groups[ trim( $parameters[3]['raw'], '"\'' ) ] ) ) {
+			$this->phpcsFile->addError( 'Please do not use cache group %s, as it is already in use by wp-memcached: https://docs.wpvip.com/technical-references/caching/object-cache/.', $stackPtr, 'Wp_memcached', $parameters[3]['raw'] );
+		}
```
jrfnl, line 97: This part needs revisiting.

Let me try and list the issues:
1. Why use `count( $parameters )` when you can use `isset( $parameters[3] )` ? (performance)
2. The `raw` index should rarely be used. If it is used at all, it should generally be combined with a token type check and a non-empty token count check.
    Consider the following code sample which would now also be flagged (but should be):
    ```php
    wp_cache_add( $key, $data, posts ); // OK.
    ```
    (while it may not be common for constants to be lowercase, it is perfectly valid PHP)
    Also consider the following which would now **not** be flagged:
    ```php
    wp_cache_add(
        $cache_key, // Key.
        'data', // Data.
        'posts' // Group name.
    ); // Error.
    ```
    I'd recommend walking the tokens in the parameter instead. Basically, you are looking for text strings and safe for a few exception, if anything else is found, you can bow out without doing the text content comparison.
    Here are some additional tests to consider:
    ```php
    // OK (undetermined).
    wp_cache_add( $key, $data, $name );
    wp_cache_add( $key, $data, 'category' . '_relationships' );
    wp_cache_add( $key, $data, "category_$relationships" );

    // These should be flagged.
    wp_cache_add( $cache_key, 'data', <<<EOD
    posts
    EOD
    );

    wp_cache_add( $cache_key, 'data', <<<'EOD'
    posts
    EOD
    );
    ```
    Again, while those last two are probably very rare, the sniff should be able to handle these.
3. Stripping quotes via `trim()` is risky. I've explained this elsewhere before. Please use the `Sniff::strip_quotes()` function instead.
4. What about using a "bow out early" pattern instead of crammed everything into one condition ? (code readability)
5. Not sure if the error code is very descriptive. What about `MemcachedGroupNameFound` or just plain `Found` ?
6. The `$data` parameter in the `addError()` method expects an array. A single text string is being passed and will cause a fatal error when this sniff is run on PHP 8.0.

Nitpick: for code readability, it would be better for the `addError()` function call to use multi-line format.

WordPressVIPMinimum/Tests/Compatibility/RestrictedCacheGroupUnitTest.inc
```diff
@@ -0,0 +1,29 @@
...
+wp_cache_set( 'cache_key', $data, "userlogins" );
+wp_cache_add( $cache_key, 'data', 'posts' );
+wp_cache_add( $cache_key,   $data,   "category_relationships", 600000 );
+wp_cache_add(
+	$key,
+	$data,
+	'usermeta'
+);
+wp_cache_set( 'key',
+// Random comment to throw things off but notrly.
```
jrfnl, line 16:
```suggestion
// Random comment to throw things off but not really.
```

WordPressVIPMinimum/Tests/Compatibility/RestrictedCacheGroupUnitTest.inc
```diff
@@ -0,0 +1,29 @@
...
+wp_cache_set( $key, [ 1, 2, 3 ],        'post_meta' );
+wp_cache_add($cache, "1", 'themes');
+
+// OK
+$key = 'cache_key';
+wp_cache_set( $key, $data, 'users_stuff' );
+wp_cache_add( $key, 'data', "posts_" );
+wp_cache_set( $key, $data, 'custom_taxonomy_relationships' );
+wp_cache_set( 'key', 'data', $userslugs );
+wp_cache_add( $key, 'post_meta', 'group' );
```
jrfnl, line 29: What about adding some test cases with partially qualified/fully qualified function and method calls ?
This should make sure that if the parent sniff is changed (when moving to PHPCSUtils), we safeguard that that doesn't break anything.

```php
\wp_cache_set( 'cache_key', $data, "userlogins" ); // Error.
namespace\wp_cache_set( 'cache_key', $data, "userlogins" ); // Ignore.
MyNamespace\wp_cache_set( 'cache_key', $data, "userlogins" ); // Ignore.
$myOwnObj->wp_cache_set( 'cache_key', $data, "userlogins" ); // Ignore.
ClassName::wp_cache_set( 'cache_key', $data, "userlogins" ); // Ignore.
```


Nitpick: All files should end on a new line character.

WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @package VIPCS\WordPressVIPMinimum
+ */
+
+namespace WordPressVIPMinimum\Sniffs\Compatibility;
+
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+
+/**
+ * This sniff checks if global memcached group names are being used.
```
jrfnl, line 13: Could this description be made more "descriptive" ? What are "global" memcached group names ?

```suggestion
 * Check if cache group names are being used which are already in use by the wp-memcached plugin which is always enabled on the VIP platform.
```

(not sure if the above is correct, but just giving you some idea of what I mean by my remark)

jrfnl (changes requested):
@rebeccahum Nice start for this sniff and using the correct abstract sniff to base it on.

I've left my remarks in-line for you to consider as this sniff does need more work.

jrfnl:
Also:

Just noticed that [the list in your comment in the originating issue](https://github.com/Automattic/VIP-Coding-Standards/issues/260#issuecomment-657648103) also includes:
```
{$taxonomy}_relationships
{$meta_type}_meta
{$taxonomy}_relationships
```

Those can be handled too (with less precision, but still), but that would require my remarks above to be addressed in full and then some special handling to be added for double quoted strings/heredocs.

jrfnl:
Oh cricky - still forgot to mention one more thing in my review: should a test for this new sniff be added to the `ruleset-test.inc` file ?

rebeccahum:
For:

```
{$taxonomy}_relationships
{$meta_type}_meta
```

I think it would be worth adding the default WP ones (e.g. `category_relationships`, `post_tag_relationships`) and then having a public property that users can set to potentially add custom taxonomies and meta?

Or would you rather have something less precise that just looks for the `_relationships` and `_meta` at the end with a manual inspection error message?

WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
```diff
@@ -0,0 +1,134 @@
...
+		'wp_cache_add'     => true,
+		'wp_cache_set'     => true,
+		'wp_cache_replace' => true,
+	];
+
+	/**
+	 * List of cache group names already in use by wp-memcached.
+	 *
+	 * @link https://github.com/Automattic/wp-memcached/blob/master/readme.txt
+	 * @var array
```
jrfnl, line 45: There should be a blank line between these two tags.

WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
```diff
@@ -0,0 +1,134 @@
...
+ */
+
+namespace WordPressVIPMinimum\Sniffs\Compatibility;
+
+use WordPressCS\WordPress\AbstractFunctionParameterSniff;
+use PHP_CodeSniffer\Util\Tokens;
+
+/**
+ * Check if cache group name is already in use by wp-memcached plugin to avoid naming collision.
+ * Note: wp-memcached is automatically enabled on WordPress VIP.
```
jrfnl, line 15: There should be a blank line between the "summary" and the "long description/additional notes".

WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
```diff
@@ -0,0 +1,134 @@
...
+		$is_concat = $this->phpcsFile->findNext(
+			Tokens::$emptyTokens,
+			$group + 1,
+			$scope_end,
+			true
+		);
+
+		if ( $this->tokens[ $is_concat ]['code'] === T_STRING_CONCAT ) {
+			return; // Bail, indeterminable.
+		}
```
jrfnl, line 121: What are you trying to do here ? Or rather: what _should_ the sniff do here ?

Please re-read my previous comment about the logic needed. This feels like you are trying to take a shortcut instead of _walking the tokens_, which is what is needed.

Think:
```php
wp_cache_set( $key, [ 1, 2, 3 ],  $prefix . 'post_meta' );
wp_cache_set( $key, [ 1, 2, 3 ],  'post_meta' + $nr );
```

WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
```diff
@@ -0,0 +1,134 @@
...
+
+		$content = $this->strip_quotes( trim( $this->tokens[ $group ]['content'] ) );
```
jrfnl, line 123: Why are you using `trim()` here ?

jrfnl, line 123: `strip_quotes()` should only be used for text strings which are quoted.

WordPressVIPMinimum/Tests/Compatibility/RestrictedCacheGroupUnitTest.inc
```diff
@@ -0,0 +1,55 @@
...
+wp_cache_add( $key, $data, posts );
+wp_cache_add( $key, $data, $name );
+wp_cache_add( $key, $data, 'category' . '_relationships' ); // Indeterminable.
+wp_cache_add( $key, $data, "category_$relationships" );
+namespace\wp_cache_set( 'cache_key', $data, "userlogins" );
```
jrfnl, line 52: These should be marked as "not the WP native function: ignore".

jrfnl (changes requested):
Done a second review. The sniff is still taking shortcuts instead of walking the tokens.

jrfnl (commented):
@rebeccahum Looking good!

Only a few tiny remarks left in-line and one question.

The question is: 
If I run the sniff over Core, I find 79 instances of the names this sniff searches for being used.
So is this sniff only about WP Memcache ? or is the sniff really about the cache groups being used by Core, which are optimized via WP Memcache ?

Depending on the answer to that question, I'd suggest possibly clarifying this a little more in the sniff class docblock and possibly in the error message thrown by the sniff.

Other than that: all good 💯

WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
```diff
@@ -0,0 +1,154 @@
...
+	 * @var string
+	 */
+	protected $group_name = 'wp_cache_functions';
+
+	/**
+	 * Functions that can assign or modify cache key.
+	 *
+	 * @var array <string function name> => <bool (true)>
+	 */
+	protected $target_functions = [
```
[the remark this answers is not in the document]
rebeccahum: @jrfnl It should include it since we don't want users touching that cache group at all. However, since `wp_cache_delete()` uses a different parameter to check for, should I modify the logic to account for that if I add that function to the list?

rebeccahum:
> So is this sniff only about WP Memcache ? or is the sniff really about the cache groups being used by Core, which are optimized via WP Memcache ?

This sniff is focusing on the global cache groups already used by core that wp-memcached extends for persistence — so it would seem the latter.

WordPressVIPMinimum/Sniffs/Compatibility/RestrictedCacheGroupSniff.php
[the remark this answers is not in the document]
jrfnl: You could change the value in the array to contain the parameter position instead of `true` and yes, then modify the sniff to use the position number to find the right parameter to examine.

I think there should be some sniffs around (either in VIPCS or in WPCS) which already do so if you'd like to look at a code pattern for this.

GaryJones:
@rebeccahum Maybe there's some use of PHPCSUtils (https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Utils-PassedParameters.html ?) to be had here now?

jrfnl:
@GaryJones As #260 has now been closed, should this PR be closed too ?

PR #624, opened 2021-02-23
ProperEscapingFunction: Flag when esc_attr is being used outside of HTML attributes

### LINKED ISSUE #155: Catch incorrect usage of esc_attr__()
For instance:

```
 <a href="..."><?php echo esc_attr__( 'testing', 'mytheme' ); ?> </a>
```

rebeccahum:
Fixes #155.

Also replaced "NOK" in the test with "Error" since "NOK" is harder to distinguish from "OK".

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
@@ -86,7 +86,7 @@ class ProperEscapingFunctionSniff extends Sniff {
 
 		$content = $this->tokens[ $html ]['content'];
 
-		if ( $this->tokens[ $html ]['code'] !== 'T_INLINE_HTML' ) {
+		if ( isset( Tokens::$stringTokens[ $this->tokens[ $html ]['code'] ] ) === true ) {
 			$content = Sniff::strip_quotes( $content );
 		}
 
--- a/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
@@ -36,11 +36,27 @@ class ProperEscapingFunctionSniff extends Sniff {
 	];
 
 	/**
+	 * List of tokens we can skip.
+	 *
+	 * @var array
+	 */
+	private $echo_or_concat_tokens =
+	[
+		T_ECHO               => T_ECHO,
+		T_OPEN_TAG           => T_OPEN_TAG,
+		T_OPEN_TAG_WITH_ECHO => T_OPEN_TAG_WITH_ECHO,
+		T_STRING_CONCAT      => T_STRING_CONCAT,
+		T_COMMA              => T_COMMA,
+	];
+
+	/**
 	 * Returns an array of tokens this test wants to listen for.
 	 *
 	 * @return array
 	 */
 	public function register() {
+		$this->echo_or_concat_tokens += Tokens::$emptyTokens;
+
 		return [ T_STRING ];
 	}
 
@@ -59,36 +75,17 @@ class ProperEscapingFunctionSniff extends Sniff {
 
 		$function_name = $this->tokens[ $stackPtr ]['content'];
 
+		$html = $this->phpcsFile->findPrevious( $this->echo_or_concat_tokens, $stackPtr - 1, null, true );
+
 		$data = [ $function_name ];
 
-		$echo_or_concat_or_html = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $stackPtr - 1, null, true );
+		if ( isset( Tokens::$textStringTokens[ $this->tokens[ $html ]['code'] ] ) === false ) { // Use $textStringTokens b/c heredoc and nowdoc tokens shouldn't be matched anyways.
+			return;
+		}
 
-		if ( $this->tokens[ $echo_or_concat_or_html ]['code'] === T_ECHO ) {
-			// Very likely inline HTML with <?php tag.
-			$php_open = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $echo_or_concat_or_html - 1, null, true );
-
-			if ( $this->tokens[ $php_open ]['code'] !== T_OPEN_TAG ) {
-				return;
-			}
-
-			$html = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $php_open - 1, null, true );
-
-			if ( $this->tokens[ $html ]['code'] === T_INLINE_HTML && $this->is_outside_html_attr_context( $function_name, $this->tokens[ $html ]['content'] ) ) {
-				$message = 'Wrong escaping function, please do not use `%s()` in a context outside of HTML attributes.';
-				$this->phpcsFile->addError( $message, $html, 'notAttrEscAttr', $data );
-				return;
-			}
-		} elseif ( $this->tokens[ $echo_or_concat_or_html ]['code'] === T_STRING_CONCAT || $this->tokens[ $echo_or_concat_or_html ]['code'] === T_COMMA ) {
-			// Very likely string concatenation mixing strings and functions/variables.
-			$html = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $echo_or_concat_or_html - 1, null, true );
-
-			if ( $this->tokens[ $html ]['code'] === T_CONSTANT_ENCAPSED_STRING && $this->is_outside_html_attr_context( $function_name, trim( $this->tokens[ $html ]['content'], '"\'' ) ) ) {
-				$message = 'Wrong escaping function, please do not use `%s()` in a context outside of HTML attributes.';
-				$this->phpcsFile->addError( $message, $html, 'notAttrEscAttr', $data );
-				return;
-			}
-		} else {
-			// Neither - bailing.
+		if ( $this->is_outside_html_attr_context( $function_name, Sniff::strip_quotes( $this->tokens[ $html ]['content'] ) ) ) {
+			$message = 'Wrong escaping function, please do not use `%s()` in a context outside of HTML attributes.';
+			$this->phpcsFile->addError( $message, $html, 'notAttrEscAttr', $data );
 			return;
 		}
 
@@ -97,6 +94,7 @@ class ProperEscapingFunctionSniff extends Sniff {
 			$this->phpcsFile->addError( $message, $stackPtr, 'hrefSrcEscUrl', $data );
 			return;
 		}
+
 		if ( $function_name === 'esc_html' && $this->is_html_attr( $this->tokens[ $html ]['content'] ) ) {
 			$message = 'Wrong escaping function. HTML attributes should be escaped by `esc_attr()`, not by `%s()`.';
 			$this->phpcsFile->addError( $message, $stackPtr, 'htmlAttrNotByEscHTML', $data );
@@ -167,12 +165,12 @@ class ProperEscapingFunctionSniff extends Sniff {
 	/**
 	 * Tests whether provided string ends with closing HTML tag in an attribute context.
 	 *
-	 * @param string $function_name Escaping attribute function name.
-	 * @param string $content Haystack in which we look for open HTML tag.
+	 * @param string $function_name Name of function found.
+	 * @param string $content       Haystack in which we look for open HTML tag.
 	 *
 	 * @return bool True if string ends with open HTML tag.
 	 */
 	public function is_outside_html_attr_context( $function_name, $content ) {
-		return $this->escaping_functions[ $function_name ] === 'attr' && substr( $content, -1 ) === '>';
+		return $this->escaping_functions[ $function_name ] === 'attr' && substr( trim( $content ), -1 ) === '>';
 	}
 }
--- a/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
@@ -23,7 +23,7 @@ class ProperEscapingFunctionSniff extends Sniff {
 	 *
 	 * @var array
 	 */
-	public $escaping_functions = [
+	protected $escaping_functions = [
 		'esc_url'    => 'url',
 		'esc_attr'   => 'attr',
 		'esc_attr__' => 'attr',
--- a/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
@@ -94,7 +94,9 @@ class ProperEscapingFunctionSniff extends Sniff {
 	 */
 	public function process_token( $stackPtr ) {
 
-		if ( isset( $this->escaping_functions[ $this->tokens[ $stackPtr ]['content'] ] ) === false ) {
+		$function_name = strtolower( $this->tokens[ $stackPtr ]['content'] );
+
+		if ( isset( $this->escaping_functions[ $function_name ] ) === false ) {
 			return;
 		}
 
@@ -104,8 +106,6 @@ class ProperEscapingFunctionSniff extends Sniff {
 		if ( $html === false || isset( Tokens::$textStringTokens[ $this->tokens[ $html ]['code'] ] ) === false ) {
 			return;
 		}
-
-		$function_name = $this->tokens[ $stackPtr ]['content'];
 
 		$data = [ $function_name ];
 
--- a/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
@@ -75,7 +75,7 @@ class ProperEscapingFunctionSniff extends Sniff {
 
 		$html = $this->phpcsFile->findPrevious( $this->echo_or_concat_tokens, $stackPtr - 1, null, true );
 
-		// Use $textStringTokens b/c heredoc and nowdoc tokens shouldn't be matched anyways.
+		// Use $textStringTokens b/c heredoc and nowdoc tokens will never be encountered in this context anyways.
 		if ( $html === false || isset( Tokens::$textStringTokens[ $this->tokens[ $html ]['code'] ] ) === false ) {
 			return;
 		}
--- a/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
@@ -73,29 +73,36 @@ class ProperEscapingFunctionSniff extends Sniff {
 			return;
 		}
 
-		$function_name = $this->tokens[ $stackPtr ]['content'];
-
 		$html = $this->phpcsFile->findPrevious( $this->echo_or_concat_tokens, $stackPtr - 1, null, true );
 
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
```diff
@@ -26,6 +26,7 @@ class ProperEscapingFunctionSniff extends Sniff {
 	public $escaping_functions = [
 		'esc_url',
 		'esc_attr',
+		'esc_attr__',
```
jrfnl, line 29: Why add `esc_attr__()` ? And if this should be added, why not the other output escaping functions ?
See: https://developer.wordpress.org/plugins/security/securing-output/

Might also be helpful to use the function names as keys and some form of grouping as the values.

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
```diff
@@ -53,8 +54,27 @@ public function process_token( $stackPtr ) {
 
 		$function_name = $this->tokens[ $stackPtr ]['content'];
 
-		$echo_or_string_concat = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $stackPtr - 1, null, true );
+		$data = [ $function_name ];
 
+		if ( $function_name === 'esc_attr' || $function_name === 'esc_attr__' ) {
+			// Find esc_attr being used between HTML tags.
+			$tokens = array_merge(
```
jrfnl, line 61: Why are you trying to _sort of_ do the same thing were as the code block below is already doing ?

Is there a specific reason why this check has to be done ahead of the rest of the checks ?

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
```diff
@@ -53,8 +54,27 @@ public function process_token( $stackPtr ) {
...
+				Tokens::$emptyTokens,
+				[
+					'T_ECHO'     => T_ECHO,
+					'T_OPEN_TAG' => T_OPEN_TAG,
+				]
+			);
+			$html_tag = $this->phpcsFile->findPrevious( $tokens, $stackPtr - 1, null, true );
+			if ( $this->tokens[ $html_tag ]['type'] === 'T_INLINE_HTML' && false !== strpos( $this->tokens[ $html_tag ]['content'], '>' ) ) {
```
jrfnl, line 69: * Why only look at inline HTML and not at other text strings, like `T_CONSTANT_ENCAPSED_STRING` ?
    Consider the following code:
    ```php
    echo '<h1>' . esc_attr__( $title, 'domain' ) . '</h1>';
    echo '<h1>', esc_attr__( $title, 'domain' ), '</h1>';
    ```
    Note: the second example passes comma separated arguments to `echo`. This is valid PHP (and more performant than concatenation).
* Is this `strpos()` correct ? Is `strpos()` the right function to use ?
    Consider the following code:
    ```php
    <h1><span class="<?php echo esc_attr__( $title, 'domain' ); ?>">Text</span></h1>
    ```

Nitpick: we moved away from Yoda conditions, Please don't re-introduce them.

jrfnl (commented):
Hi Rebecca, left some questions again for you to consider.

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
rebeccahum: I figured the other ones could be added in a separate PR since it doesn't relate to the `esc_attr*()` logic being modified here specifically.

> Might also be helpful to use the function names as keys and some form of grouping as the values.

Noted!

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
rebeccahum: I think `substr()` would be a better one to use over `strpos()` since we only care about the `>` being on the last part of the stack pointer content (if it's a `T_CONSTANT_ENCAPSED_STRING` we might have to just strip out any trailing quotes).

The Yoda conditions are a bad habit, must have sneaked in unwillingly 😄.

TIL about `,` being used as a concatenation operator!

rebeccahum:
@jrfnl Thank you for the useful feedback, I've iterated on it! 🙂

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
rebeccahum: @jrfnl Thank you for the useful feedback, I've iterated on it! 🙂

Why only look at inline HTML and not at other text strings, like T_CONSTANT_ENCAPSED_STRING ?

Do you think I need to expand the sniff to account for more text string tokens (besides T_CONSTANT_ENCAPSED_STRING) such as the ones listed in $textStringTokens? https://github.com/squizlabs/PHP_CodeSniffer/blob/8b0137a78cfae9b8fe1c8afdf5396c8a8d6a25af/src/Util/Tokens.php#L537?

jrfnl: > TIL about , being used as a concatenation operator!

Sorry for confusing you. The `,` is *not* a concatenation operator. `echo` as a language construct, however, allows for passing multiple "arguments" which will then each be echo-ed out. The net effect on the output is the same, but it can save a lot of memory using comma's instead of concatenation.

As per the [documentation](https://www.php.net/manual/en/function.echo.php):
>  Outputs all parameters. No additional newline is appended.
> 
> `echo` is not actually a function (it is a language construct), so you are not required to use parentheses with it. `echo` (unlike some other language constructs) does not behave like a function, so it cannot always be used in the context of a function. Additionally, if you want to pass more than one parameter to `echo`, the parameters must not be enclosed within parentheses. 

Note: this only works for `echo`, not for `print`.

jrfnl: > I think substr() would be a better one to use over strpos() since we only care about the `>` being on the last part of the stack pointer content (if it's a T_CONSTANT_ENCAPSED_STRING we might have to just strip out any trailing quotes).

Sounds like a better approach.

> Do you think I need to expand the sniff to account for more text string tokens (besides T_CONSTANT_ENCAPSED_STRING) such as the ones listed in $textStringTokens? 

Well, heredoc and nowdoc do not allow for function calls within the text block. Even so, that shouldn't be an issue as that just means those tokens would (should) never match. So, yes, using `Tokens::$textStringTokens` will probably be a better approach.

jrfnl:
@rebeccahum Shall I continue with the questions based on the new code or shall we talk it through in a call ?

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
```diff
@@ -47,41 +53,45 @@ public function register() {
...
+			$php_open = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $echo_or_concat_or_html - 1, null, true );
 
 			if ( $this->tokens[ $php_open ]['code'] !== T_OPEN_TAG ) {
 				return;
 			}
 
 			$html = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $php_open - 1, null, true );
 
-			if ( $this->tokens[ $html ]['code'] !== T_INLINE_HTML ) {
+			if ( $this->tokens[ $html ]['code'] === T_INLINE_HTML && $this->is_outside_html_attr_context( $function_name, $this->tokens[ $html ]['content'] ) ) {
```
jrfnl, line 76: * Why are the checks now (in duplicate) in the if/else ? Why can't this check be moved further down to whether the other checks are ?
* What happens now if the token is not `T_INLINE_HTML` (or in the case below `T_CONSTANT_ENCAPSED_STRING`) ? Is this intentional ?

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
```diff
@@ -47,41 +53,45 @@ public function register() {
...
+				$message = 'Wrong escaping function, please do not use `%s()` in a context outside of HTML attributes.';
+				$this->phpcsFile->addError( $message, $html, 'notAttrEscAttr', $data );
 				return;
 			}
-		} elseif ( $this->tokens[ $echo_or_string_concat ]['code'] === T_STRING_CONCAT ) {
+		} elseif ( $this->tokens[ $echo_or_concat_or_html ]['code'] === T_STRING_CONCAT || $this->tokens[ $echo_or_concat_or_html ]['code'] === T_COMMA ) {
```
jrfnl, line 81: * Did you know you can use the WPCS [`Sniff::strip_quotes()`](https://github.com/WordPress/WordPress-Coding-Standards/blob/7da1894633f168fe244afc6de00d141f27517b62/WordPress/Sniff.php#L950) method for more reliably quote stripping ? (will be replaced in WPCS 3.0 by the PHPCSUtils `TextStrings::stripQuotes()` method which does the same)

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
```diff
@@ -24,9 +24,15 @@ class ProperEscapingFunctionSniff extends Sniff {
...
+		'esc_url'    => 'url',
+		'esc_attr'   => 'attr',
+		'esc_attr__' => 'attr',
+		'esc_attr_x' => 'attr',
+		'esc_attr_e' => 'attr',
+		'esc_html'   => 'html',
+		'esc_html__' => 'html',
+		'esc_html_x' => 'html',
+		'esc_html_e' => 'html',
 	];
```
jrfnl, line 36: Have you considered code like this ?
```php
<div><?= esc_attr($attr) ?></div><!-- Error -->
<div><?php esc_attr_e($attr) ?></div><!-- Error -->
```

(Sniff failure on this code is only partially due to the additional functions being added)

jrfnl (commented):
Next round of questions for now, we can talk them through in a call next week if you like.

rebeccahum:
Ready for another round, @jrfnl!

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
```diff
@@ -47,46 +69,32 @@ public function register() {
...
-			// Very likely string concatenation mixing strings and functions/variables.
-			$html = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $echo_or_string_concat - 1, null, true );
+		$data = [ $function_name ];
 
-			if ( $this->tokens[ $html ]['code'] !== T_CONSTANT_ENCAPSED_STRING ) {
-				return;
-			}
-		} else {
-			// Neither - bailing.
+		if ( isset( Tokens::$textStringTokens[ $this->tokens[ $html ]['code'] ] ) === false ) { // Use $textStringTokens b/c heredoc and nowdoc tokens shouldn't be matched anyways.
```
jrfnl, line 82: This line with the trailing comment has become quite long. Maybe put the comment on the line above it ?

jrfnl (commented):
Hi Rebecca, as discussed in the call: well done! This is looking really good!

Just some nitpicks and minor cleanup left to do, but other than that, ready to merge!

General remark regarding the tests: I'm missing tests with a `T_DOUBLE_QUOTED_STRING` token before the call to `esc_*()`.

Not for this PR, but for a possible future iteration: using `strip_quotes()` for all three checks and only when the token is a single/double quoted string.

rebeccahum:
@jrfnl Besides having a hard time kicking the yoda habit, I think this should be good for one more review!

WordPressVIPMinimum/Tests/Security/ProperEscapingFunctionUnitTest.inc
```diff
@@ -30,12 +30,44 @@ echo "<a title='" . esc_html( $some_var ) . "'></a>"; // NOK.
...
 echo 'data-param-url="' . esc_url( $share_url ) . '"'; // OK.
 
-echo 'data-param-url="' . esc_html( $share_url ) . '"'; // NOK.
+echo 'data-param-url="' . esc_html( $share_url ) . '"'; // Error.
 
 ?>
 
 <form method="post" action="<?php echo esc_html(admin_url('admin.php?page='.$base_name.'&amp;mode=logs&amp;id='.$poll_id)); ?>">
+
+?>
```
jrfnl, line 43: You're not in PHP at this point, so this `?>` is superfluous (and can cause the test code to be tokenized incorrectly).

WordPressVIPMinimum/Tests/Security/ProperEscapingFunctionUnitTest.inc
```diff
@@ -30,12 +30,44 @@ echo "<a title='" . esc_html( $some_var ) . "'></a>"; // NOK.
...
+<h1><?php echo esc_attr__( $title, 'domain' ); ?></h1> <!-- Error --> ?>
+<?php echo '<h1>' . esc_attr__( $some_var, 'domain' ) . '</h1>'; // Error.
+echo '<h1>',      esc_attr__( $title, 'domain' ), '</h1>'; // Error.
+echo "<" . "h1>" . esc_attr( $test ) . '</h1>'; // Error.
+?>
+<h1> <?php echo esc_attr( $title ) . '</h1>'; ?> // Error.
+<div><?= esc_attr($attr) ?></div><!-- Error -->
+<div><?php esc_attr_e($attr) ?></div><!-- Error -->
+<div data-tag="<?= esc_attr( $attr ); ?>"> <!-- OK -->
+<?php echo "<div> {$test} </div>"; // OK.
```
jrfnl, line 71: What is this line testing ?

WordPressVIPMinimum/Tests/Security/ProperEscapingFunctionUnitTest.inc
```diff
@@ -30,12 +30,44 @@ echo "<a title='" . esc_html( $some_var ) . "'></a>"; // NOK.
...
+echo "<{$tag}>" . esc_attr( $tag_content ) . "</{$tag}>"; // Error.
+echo "<$tag" . '   >' . esc_attr( $tag_content ) . "</$tag>"; // Error.
```
jrfnl, line 73: What is this line testing ?

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
```diff
@@ -47,47 +69,40 @@ public function register() {
...
+		$content = $this->tokens[ $html ]['content'];
 
-			if ( $this->tokens[ $html ]['code'] !== T_INLINE_HTML ) {
-				return;
-			}
-		} elseif ( $this->tokens[ $echo_or_string_concat ]['code'] === T_STRING_CONCAT ) {
-			// Very likely string concatenation mixing strings and functions/variables.
-			$html = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $echo_or_string_concat - 1, null, true );
+		if ( isset( Tokens::$stringTokens[ $this->tokens[ $html ]['code'] ] ) === true ) {
+			$content = Sniff::strip_quotes( $content );
```
jrfnl, line 90: 👍 

The only thing is, now this has been changed, the code in the `attr_expects_url()` and the `is_html_attr()` functions needs to be updated too.

The fact that the tests aren't failing on it, only highlights the fact that tests for some parts of those functions are missing (or more likely even: that the author of the sniff couldn't get the tests working as some of the code in those functions is actually faulty).

Suggested unit tests:
```php
// These should throw the `htmlAttrNotByEscHTML` error.
echo '<div class=\'' . esc_html($class) . '\'>'; // Error.
echo "<div class=\"" . esc_html($class) . '">'; // Error.
echo "<div $someAttribute class=\"" . esc_html($class) . '">'; // Error.

// These should throw the `hrefSrcEscUrl` error.
echo '<a href=\'' . esc_html($url) . '\'>'; // Error.
echo "<img src=\"" . esc_html($src) . '"/>'; // Error.
echo "<div $someAttributeName-url=\"" . esc_html($url) . '">'; // Error.
```

jrfnl (dismissed):
@rebeccahum We're inching closer and closer.

> Not for this PR, but for a possible future iteration: using strip_quotes() for all three checks and only when the token is a single/double quoted string.

As you've started to address this remark within this PR now anyway, it, unfortunately, has opened up the road to yet more changes which will be needed to finish that refactor off properly.

Let me know if you want to talk my current remarks through in another call. Or even if you'd want me to add a commit to the PR to address that last part for the `attr_expects_url()` and the `is_html_attr()` functions.

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
jrfnl: Also:

```php
echo '<a href=', esc_html($url), '>';
```

WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
```diff
@@ -47,47 +69,40 @@ public function register() {
...
-		if ( in_array( $this->tokens[ $stackPtr ]['content'], $this->escaping_functions, true ) === false ) {
+		if ( isset( $this->escaping_functions[ $this->tokens[ $stackPtr ]['content'] ] ) === false ) {
 			return;
 		}
 
-		$function_name = $this->tokens[ $stackPtr ]['content'];
+		$html = $this->phpcsFile->findPrevious( $this->echo_or_concat_tokens, $stackPtr - 1, null, true );
 
-		$echo_or_string_concat = $this->phpcsFile->findPrevious( Tokens::$emptyTokens, $stackPtr - 1, null, true );
+		// Use $textStringTokens b/c heredoc and nowdoc tokens shouldn't be matched anyways.
```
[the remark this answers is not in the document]
jrfnl: Don't know where those `<span>`'s are coming from... and can't see them in the edit view to remove either....

WordPressVIPMinimum/Tests/Security/ProperEscapingFunctionUnitTest.inc
rebeccahum: This line is testing that there can be a concatenated start tag before `esc_attr()` is called will still flag an error.

rebeccahum: It doesn't test anything new but it ensures that this use case will not be broken if the sniff is changed in the future.

jrfnl (approved):
Approving with the addition of the seven extra commits which we discussed in a videocall.
See the commit messages of each of these commits for details.

One final note:
The _format_ of the `public` property `$escaping_functions` has changed - from the "function names as values" to "function names as keys with their type as value".

This format change **is a breaking change** as the sniff will no longer function correctly if the old property format is used in a custom ruleset.

As discussed in our call and documented in the commit message of the last commit:

> This is a breaking change, but as:
> 1. The property was never intended to be changed from a custom ruleset and
> 2. The chances of any custom rulesets actually existing which _do_ change this property being very, very small...
> 
> ...I believe it is justified to make the breaking change in the format in a minor release as the format change provides a real functional benefit.

With that in mind, the property has now also been made `protected`.

Deprecating the `public` property and setting up a complex array merge mechanism with the "new" `protected` property just in case there is a custom ruleset which _does_ overrule the value seems overkill as the chances of this actually breaking anything in real life are so small.

This breaking change, however, will need a very clear entry in the changelog on the off-chance of a custom ruleset using the property.

/cc @GaryJones

PR #626, opened 2021-02-24
IncludingFile: Add custom keywords to detect to reduce false positives

### LINKED ISSUE #407: Minimize false positives on WordPressVIPMinimum.Files.IncludingFile.IncludingFile
`require_once \PL\PATH . '/inc/endpoints/especiales-endpoint.php';`

`❗️ Warning: Absolute include path must be used. Use get_template_directory(), get_stylesheet_directory() or plugin_dir_path() (WordPressVIPMinimum.Files.IncludingFile.IncludingFile).`

Currently clients include files be defining path variables or constants that get used for the path. This makes the IncludingFile rule very noisy with false postiives.

We should enhance the rule check to account for path variables/constants to reduce false positives.

rebeccahum:
Fixes #407.

Takes into account for URLs and $restrictedConstants with keywords from $customPaths.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
@@ -56,6 +56,17 @@ class IncludingFileSniff extends AbstractFunctionRestrictionsSniff {
 	];
 
 	/**
+	 * List of custom keywords for paths.
+	 *
+	 * @var array
+	 */
+	public $customPaths = [
+		'path',
+		'dir',
+		'base',
+	];
+
+	/**
 	 * Functions used for modify slashes.
 	 *
 	 * @var array
@@ -102,6 +113,16 @@ class IncludingFileSniff extends AbstractFunctionRestrictionsSniff {
 		if ( $this->tokens[ $nextToken ]['code'] === T_DIR || $this->tokens[ $nextToken ]['content'] === '__DIR__' ) {
 			// The construct is using __DIR__ which is fine.
 			return;
+		}
+
+		if ( $this->tokens[ $nextToken ]['code'] === T_CONSTANT_ENCAPSED_STRING && filter_var( str_replace( [ '"', "'" ], '', $this->tokens[ $nextToken ]['content'] ), FILTER_VALIDATE_URL ) ) {
+			$message = 'Include path must be local file source, external URLs are prohibited on WordPress VIP.';
+			$this->phpcsFile->addError( $message, $nextToken, 'ExternalURL' );
+			return;
+		}
+
+		if ( $this->has_custom_path( $nextToken ) !== false && array_key_exists( $this->tokens[ $nextToken ]['content'], $this->restrictedConstants ) === false ) {
+			return; // Bail since it has a keyword from $customPaths but not in list of $restrictedConstants.
 		}
 
 		if ( $this->tokens[ $nextToken ]['code'] === T_VARIABLE ) {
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -104,6 +115,16 @@ public function process_token( $stackPtr ) {
 			return;
 		}
 
+		if ( $this->tokens[ $nextToken ]['code'] === T_CONSTANT_ENCAPSED_STRING && filter_var( str_replace( [ '"', "'" ], '', $this->tokens[ $nextToken ]['content'] ), FILTER_VALIDATE_URL ) ) {
```
jrfnl, line 118: Why did you move this check up ?

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -104,6 +115,16 @@ public function process_token( $stackPtr ) {
...
+			$message = 'Include path must be local file source, external URLs are prohibited on WordPress VIP.';
+			$this->phpcsFile->addError( $message, $nextToken, 'ExternalURL' );
+			return;
+		}
+
+		if ( $this->has_custom_path( $nextToken ) !== false && array_key_exists( $this->tokens[ $nextToken ]['content'], $this->restrictedConstants ) === false ) {
```
jrfnl, line 124: * Why are you checking against `!== false` when the method will only ever return `true` or `false` ?
* Why use `array_key_exists()` instead of `isset()` ?
* Why are these conditions in this order ?

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
```diff
@@ -163,13 +184,35 @@ public function process_token( $stackPtr ) {
...
+	 *                         This parameter needs to be passed if no $content is
+	 *                         passed.
+	 * @param string $content  Optionally, the current content string, might be a
+	 *                         substring of the original string.
+	 *                         Defaults to `false` to distinguish between a passed
+	 *                         empty string and not passing the $content string.
+	 *
+	 * @return bool True if the string contains a keyword in $customPaths, false otherwise.
+	 */
+	public function has_custom_path( $stackPtr = null, $content = null ) {
```
jrfnl, line 205: * Why is this a `public` method ?
* Why are both parameters optional ?
* Why is no check being done against the token code/type ? This check will now also bail out on a call to `dirname()` or on the variable `$custom_path` which is not the intention of this check.

WordPressVIPMinimum/Tests/Files/IncludingFileUnitTest.inc
```diff
@@ -13,6 +13,10 @@ require dirname( __FILE__, 2 ) . '/my_file.php' );
 require dirname( dirname( __FILE__ ) ) . '/my_file.php' );
 require dirname( dirname( __DIR__ ) ) . '/my_file.php' );
 require_once( get_parent_theme_file_path( '/my_file.php' ); );
+require_once(    $custom_path );
```
jrfnl, line 16: This one should trigger the `UsingVariable` error code, but doesn't.

jrfnl (commented):
Hi Rebecca,

I've been looking at this change and notice I find it difficult to review it as there is so much more which needs improving in this sniff, that it is hard to focus on this specific change.

This might actually be a good sniff to take the time to dissect in detail in a live coding call and talk through every code inefficiency and assumption in the code which should be fixed.

For now, specifically for this PR, let's start with some questions (left in-line)...

WordPressVIPMinimum/Sniffs/Files/IncludingFileSniff.php
rebeccahum: I moved it up so that it doesn't bail before checking for an external URL with a keyword from `$customPaths` in it. e.g.:

```
include( "http://path.com/bad_file.php" );
```

rebeccahum:
Closing for now as per our discussion on making this an atomic commit for the next release and then doing more re-factoring for 3.0.

PR #633, opened 2021-03-03
CI: switch to GitHub Actions

### LINKED ISSUE #628: CI: move away from Travis-ci.org
Travis-ci.org is shutting down.

image

I know that in the past one could already manually move accounts over from .org to .com by a) signing up to .com, b) enabling the migration beta and then moving whichever repo you wanted to move.

It might be worth checking who within the Automattic organisation has the right access rights to the Travis account to do so and approach them about getting this repo moved (to prevent CI potentially having downtime when the move happens unexpectedly or when travis-ci.org is shut down without the account having been moved).

Alternatively, now might be a good time to move the CI to GitHub actions and away from Travis.

If the preference is to move to travis-ci.com, can someone from within Automattic please take ownership of this ticket and move that forward ?

If the preference is to move to GH actions, I can set that up and send in a PR.

jrfnl:
## TL;DR

This switches out the Travis script for GitHub Actions workflows.

This initial pull request is a one-on-one translation of the original CI script to GH actions.

The only differences are:
* The Travis "sniff" and "lint" stages are combined in one GH Actions workflow job.
* Any violations reported by the `xmllint` check as well as by the PHPCS check, will now be shown in-line in a PR.
* The `test` job is set up to allow for supporting multiple WPCS versions, even though VIPCS at the time of writing only supports one version (`2.3.0`).
* The build against PHP 8.1/nightly is commented out for the time being. See the description for the second commit for more details.

Fixes #628

## Some ideas for potential future iterations

* (Re-)enable a test run against PHP 8.1/nightly.
* Add a "Quick test" script to run the tests against a more limited matrix for pushes only.
* Implement use of the [PHP Parallel Lint](https://github.com/php-parallel-lint/PHP-Parallel-Lint/) package for linting the PHP code and showing any errors in-line in a PR.
* Implement use of the [PHPCS feature completeness check](https://github.com/PHPCSStandards/PHPCSDevTools/#checking-whether-all-sniffs-in-a-phpcs-standard-are-feature-complete).
* Add a test run against PHPCS 4.x when it is closer to being released.
* Register with coveralls.io (or another service) and start recording and checking code coverage for builds.

### CI: switch to GitHub Actions - step 1: sniff and lint stage

This commit:
* Adds a GH Actions workflow for the CI checks which were previously run on Travis in the `sniff` and `lint` stages.
    For the badge and workflow display, the name `BasicQA` seems an appropriately descriptive name.
* Removes that part of the `.travis.yml` configuration.
* Removes a duplicate line in the `bin/xml-lint` script used by this workflow.

Notes:
1. In the Travis script, these checks were in two separate stages. I've now combined them into one workflow `job`.
    In GH Actions, the environment has to be setup as part of the workflow. And as the environment needed for these checks is largely the same, combining the checks into one workflow simplifies the script and reduced duplication.
2. Builds will run on all pushes and on pull requests, with the exception of pushes just modifying files which are irrelevant to this workflow.
3. The workflow can be manually (re-)triggered if needs be.
4. Any violations reported by the XML validation against the XSD schema, as well as coming from the PHPCS run, are set up to be shown in-line in the pull request to easily see what line in the pull request causes a build to fail.

### CI: switch to GitHub Actions - step 2: test stage

This commit:
* Adds a GH Actions workflow for the CI check which was previously run on Travis in the `test` stage.
* Removes the, now redundant, `.travis.yml` configuration.
* Updates the `.gitattributes` file.

Notes:
1. For the time being, I've commented out the job against PHP 8.1/nightly as - no matter what I've tried - I can't even get the tests running at the moment, so they would only ever fail.
    Once PHP 8.1 starts to release alpha/beta/RC builds, further testing should be done to attempt to get this working.
    Notes/current findings:
    - With PHPUnit 7.x on PHP 8.1, using a configuration file will block the tests from running.
    - In some cases (PHPCS itself, PHPCSExtra), I managed to get the tests running by passing all relevant information via the command-line and actively ignoring the configuration file using `--no-configuration`. Unfortunately, that does not seem to work in this case. Further investigation is needed.
    - Based on the (running) tests for PHPCSExtra on PHP 8.1, an incompatibility with PHP 8.1 in PHPCS itself has already been discovered.
        A fix for this has been pulled, but not yet merged: squizlabs/php_codesniffer#3250


jrfnl:
Note: for this PR to become mergable, the "required statuses" will need to be updated.

GaryJones (approved):
Approved with one small query.

.github/workflows/test.yml
```diff
@@ -0,0 +1,101 @@
...
+            wpcs_version: '2.3.*'
+            experimental: false
+
+          # Experimental builds. These are allowed to fail.
+          #- php: '8.1'
+          #  phpcs_version: 'dev-master'
+          #  wpcs_version: '2.3.*'
+          #  experimental: true
+
+    name: "Test: PHP ${{ matrix.php }} - PHPCS ${{ matrix.phpcs_version }} - WPCS ${{ matrix.wpcs_version }}"
```
GaryJones, line 52:
```suggestion
    name: "PHP ${{ matrix.php }} - PHPCS ${{ matrix.phpcs_version }} - WPCS ${{ matrix.wpcs_version }}"
```

Each line is already prefixed with `Test` from line 1, so it seems redundant to include it again here.

Screenshot 2021-03-04 at 11 17 45

.github/workflows/basics.yml
```diff
@@ -0,0 +1,62 @@
...
+        continue-on-error: true
+        run: vendor/bin/phpcs --report-full --report-checkstyle=./phpcs-report.xml
+
+      - name: Show PHPCS results in PR
+        run: cs2pr ./phpcs-report.xml
+
+      # Validate the composer.json file.
+      # @link https://getcomposer.org/doc/03-cli.md#validate
+      - name: Validate Composer installation
+        run: composer validate --no-check-all --strict
```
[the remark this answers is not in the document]
jrfnl: Good point - I left the order as it was in the Travis script, but will move it up.

.github/workflows/test.yml
jrfnl: > Each line is already prefixed with Test from line 1, so it seems redundant to include it again here.

I realize it seems that way, however, there is a reason for it.

In the PR block below, each check is shown in the form "workflow name/job name/trigger event". This includes ALL checks, i.e. both those done on `push` as well as those run on `pull_request`. So yes, here it is easy to see what's what and the prefix looks redundant.

However, now, check in the "Settings" -> "Branches" -> protected branch/required statuses.

There you will _only_ see the "job name" to choose which jobs to set as required. For now, that is not an issue, but as soon as I would add the "Quick test" workflow as is being discussed in #628, you would no longer be able to tell which runs are from the "Quick test" job (not required) and which ones are from "Test" job (required).

Does that make it clearer why I prefixed the job names ?

.github/workflows/basics.yml
```diff
@@ -0,0 +1,62 @@
...
+    paths-ignore:
+      - '**.md'
+  pull_request:
+  # Allow manually triggering the workflow.
+  workflow_dispatch:
+
+jobs:
+  checkcs:
+    name: 'Basic CS and QA checks'
+    runs-on: ubuntu-latest
```
GaryJones, line 16: The tests have got a warning:

> Ubuntu-latest workflows will use Ubuntu-20.04 soon. For more details, see https://github.com/actions/virtual-environments/issues/1816

Presumably this won't matter as there is `shivammathur/setup-php@v2` being used for Basic and Test workflows already anyway?

GaryJones (approved):
🎉 ✅

.github/workflows/basics.yml
jrfnl: That warning is not directly related to the setup-php action. It is related to the jobs all using `runs-on: ubuntu-latest`.

For all intends and purposes, the ubuntu version shouldn't have any significant effect on the PHP environment (though, yes, that is providing the `setup-php` action will work on all), so I've been ignoring the warning.

Using `ubuntu-latest` is a shorthand way to always use the latest instead of a "fixed" version.
With Travis, there was a lot of differences between the images (`trusty`, `xenial`, `bionic`) and what they supported and it was more important to "fix" it to a specific version as it impacted which versions of PHP were available.

I don't have the impression the difference is as big on GH Actions, nor that it will impact the workflows we use.
Where it would, I'm expecting the `setup-php` action to fill in the difference.
Also see: https://github.com/shivammathur/setup-php/#cloud-osplatform-support

If and when GH actually switches over and the build would fail, we could always (temporarily) fix it to a specific `ubuntu` version until support for the next one has been repaired.

More than anything, it's about us not having to keep track of the specifics of the images and not having to update the name of the images used every so often and trusting that the actions we use will handle it.

Does that explain it well enough ?

PR #635, opened 2021-03-09
WordPress-VIP-Go: Remove WordPress.WP.GlobalVariablesOverride.Prohibited

rebeccahum:
Per @jrfnl's [recommendation](https://github.com/Automattic/VIP-Coding-Standards/pull/610#issuecomment-790186115), we should just remove the rule since it hasn't been used for awhile anyways.


rebeccahum:
Note: Tests shouldn't fail after #636 is merged.

jrfnl (changes requested):
Just to clarify: I recommended for the _severity change in the VIP-Go ruleset_ to be removed. The VIPMinimum ruleset includes the sniff and VIP-Go inherits from VIPMinimum, so once the severity change is removed, all issues will (and should) still be reported and reported at `severity` `5`.

This also means that:
* The VIP-Go ruleset test should **not** be adjusted. The error should still be thrown, just at severity 5, not 3.
* If the ruleset test currently does not receive the error, something else may be going on and would need investigating.
* As for line 300 - once the tests on line 187/188 are brought back, that line should not need to be changed.

WordPress-VIP-Go/ruleset.xml
```diff
@@ -210,10 +210,7 @@
 		<!-- We trust that translations are safe on VIP Go -->
 		<severity>1</severity>
 	</rule>
-	<rule ref="WordPress.WP.GlobalVariablesOverride.Prohibited">
-		<!-- This is often a false positive. Still nice to flag for a check -->
-		<severity>3</severity>
-	</rule>
+	<rule ref="WordPress.WP.GlobalVariablesOverride.Prohibited"/>
```
jrfnl, line 213: This line is not needed. The sniff (including this error code) is inherited from `WordPressVIPMinimum`, so doesn't have to be included explicitly.

Only reason to still have it, would be if the property I mentioned in my comment would be added: https://github.com/Automattic/VIP-Coding-Standards/pull/610#issuecomment-790186115

rebeccahum: I think I'll keep the property disabled for the time being since it seems to be discouraged per:
> Unfortunately, at this time, it is not possible to set this property for only a select group of files. With that in mind, setting this property is discouraged.
If we receive reports of it being an issue, we can re-visit later I'm thinking.

jrfnl (approved):

PR #636, opened 2021-03-09
WordPress-VIP-Go: Add ignore statement for ruleset since VariableAnalysis.CodeAnalysis.VariableAnalysis 

rebeccahum:
It was being triggered in the ruleset tests

### DIFF
--- a/WordPress-VIP-Go/ruleset-test.inc
+++ b/WordPress-VIP-Go/ruleset-test.inc
@@ -297,7 +297,7 @@ get_user_meta(); // Ok.
 update_user_meta( $bar, '123', $foo ); // Ok.
 
 // WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__usermeta
-$query = "SELECT * FROM $wpdb->usermeta"; // Ok.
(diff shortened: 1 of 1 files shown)


jrfnl:
My comment is superseded by the issue being closed while I was writing it.

Posting it anyway in case this comes up again.

---

With the current `develop` branch I cannot reproduce the issue this is supposedly solving.
Running ` phpcs ./wordpress-vip-go/ruleset-test.inc --standard=wordpress-vip-go` does not show any errors or warnings on line 300.

I might be misunderstanding, so could you please elaborate on what this is supposed to be solving ?

Other notes:
* Is an ignore comment the right way to solve it ? Or should the variable be added to the list of variables at the top of the file ?
* If an ignore comment is the way to go: can the ignore comment be made error code specific ?

rebeccahum:
@jrfnl The test was failing on #635 and I mistakenly thought it was unrelated but after further inspection, it was related so I'm putting it in that PR. I think an ignore comment is the way to go because it's testing for `WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__usermeta` and therefore, should have no dependencies to other rules.

PR #638, opened 2021-03-26
IncludingFileSniff: add property of custom keywords to not trigger sniff

rebeccahum:
We're just doing an atomic commit for this one to solve #407 in reducing false positives.

However, the entire IncludingFileSniff needs a facelift with some breaking changes, so the work started in #626 will continue.


GaryJones (approved):
Approved.

PR #644, opened 2021-03-29
RestrictedFunctionsSniff: Remove wp_cache_get_multi group

### LINKED ISSUE #642: Remove error code WordPressVIPMinimum.Functions.RestrictedFunctions.wp_cache_get_multi_wp_cache_get_multi
Now that Core has `wp_cache_get_multiple()` [available](https://developer.wordpress.org/reference/functions/wp_cache_get_multiple/), we don't need to flag for `wp_cache_get_multi()` since we are transitioning that function out of our mu-plugins.

rebeccahum:
Fixes #642.


GaryJones (approved):
We still do have customers stuck on WP 5.4, but I'm +1 with moving forward with this change now.

PR #648, opened 2021-03-31
Add new OptionsRaceConditionSniff

### LINKED ISSUE #637: Add sniff to delete_option() + add_option
As explained at https://github.com/woocommerce/woocommerce/pull/27696 there is a race condition when `delete_option('foo')` is followed by `add_option('foo', '...')`.

Having a sniff that looked for a `delete_option()` call followed by an `add_option()` immediately 
or at some point within the same scope, for the same option key, could flag this race condition and suggest using `update_option()` instead.

```php
function reset_settings() {
	$defaults = get_default_settings();
	delete_option( 'my_settings' );
	add_option( 'my_settings', $defaults );
}
```

Likely to need some consideration of an option key that is using a variable.

- `add_option()` in a different scope.
- different option key (including different variable).

rebeccahum:
Fixes #637.

Let the roasting begin, @jrfnl! 😄

### DIFF
--- a/WordPressVIPMinimum/Tests/Functions/OptionsRaceConditionUnitTest.inc
+++ b/WordPressVIPMinimum/Tests/Functions/OptionsRaceConditionUnitTest.inc
@@ -1,0 +1,112 @@
+<?php
+
+function test_with_function_between() {
+	$defaults = get_default_settings();
+	delete_option( 'my_settings' );
+
+	do_stuff();
+
+	add_option( 'my_settings', $defaults ); // OK - since there is stuff being done in between.
+}
+
+function test_different_option_name() {
+	$defaults = get_default_settings();
+	delete_option( 'my_settings' );
+	add_option( 'settings', [] ); // OK - Not same key as delete_option().
+}
+
+function test_same_option_key_by_var_assignment() {
+	$key = 'option_key';
+	delete_option( $key );
+	$add = add_option( $key, $key ); // Warning.
+}
+
+function test_same_option_key() {
+	$key = 'option_key';
+	delete_option( $key );
+	add_option( $key, 'data' ); // Warning.
+}
+
+function test_same_option_key_var_assignment2() {
+	$delete = delete_option( 'option_key' );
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Tests/Functions/OptionsRaceConditionUnitTest.inc
```diff
@@ -0,0 +1,112 @@
...
+	$test = add_option( 'test_option', [ 1, 2, 3 ] ); // Warning.
+}
+
+function concurrent_option_writes_with_one_if_same_key() {
+	$test = delete_option( 'test_option' );
+	if ( $test ) {
+		add_option( 'test_option', [] ); // Warning.
+	}
+	$test = add_option( 'test_option', [ 1, 2, 3 ] ); // Warning.
+}
```
GaryJones, line 112: How about:

```php
delete_option( 'foobar' );
add_option( 'foo' . 'bar', 'baz' );

delete_option( 'foobar' );
add_option( "foobar", 'baz' );

delete_option( 'foobar' );
add_option( '     foobar   ', 'baz' ); // Internally `add_option()` calls `trim()` on the option name.
```
and other variants of key names in different formats?

rebeccahum: I don't see this being a common case:

```
delete_option( 'foobar' );
add_option( 'foo' . 'bar', 'baz' );
```

But I could see something like this occurring:

```
delete_option( 'foo_' . $bar );
add_option( 'foo_' . $bar, 'baz' );
```

jrfnl:
@rebeccahum Instead of roasting (and sorry, yes it would be a painful roasting), let's try something different.

Let's start with a little exercise:
* Can you try to write out - in words, not code - the "code pattern" you are trying to look for ?
* Then make a list of all the requirements which need to be met for the code pattern to be considered a match.
* Next, look at the code you've written and try to find all the assumptions you've made in it to account for specific code samples instead of accounting for the pattern.

PR #649, opened 2021-03-31
RestrictedFunctionsSniff: Change default message of switch_to_blog()

### LINKED ISSUE #643: switch_to_blog: Change message to reflect WordPress-VIP-Go ruleset and downgrade to "Warning"
This is a two-part issue since I _think_ step #2 will need to be done in 3.x due to it being a breaking change (can you please confirm @jrfnl?)

**Step 1**
Mentioned in #615, we should change the message in the `WordPressVIPMinimum.Functions.RestrictedFunctions.switch_to_blog_switch_to_blog` error code to reflect the one being changed in WordPress-VIP-Go since the current one is outdated.

From:

https://github.com/Automattic/VIP-Coding-Standards/blob/6d3f5267bee6f0e7548aacc3881b2ac0bcd981e9/WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php#L103

To: https://github.com/Automattic/VIP-Coding-Standards/blob/1fa0266752aed5f53b369e3e60858f03d9ff9aac/WordPress-VIP-Go/ruleset.xml#L249

**Step 2**
Downgrade to warning type at sniff level and remove modification in `WordPress-VIP-Go` ruleset:

https://github.com/Automattic/VIP-Coding-Standards/blob/1fa0266752aed5f53b369e3e60858f03d9ff9aac/WordPress-VIP-Go/ruleset.xml#L247

rebeccahum:
Fixes #643.


WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
```diff
@@ -98,7 +98,7 @@ public function getGroups() {
 			// @link https://docs.wpvip.com/technical-references/code-review/vip-notices/#h-switch_to_blog
 			'switch_to_blog' => [
 				'type'      => 'error',
-				'message'   => '%s() is not something you should ever need to do in a VIP theme context. Instead use an API (XML-RPC, REST) to interact with other sites if needed.',
+				'message'   => '%s() may not work as expected since it only changes the database context for the blog and does not load the plugins or theme of that site. Filters or hooks on the blog you are switching to will not run.',
```
GaryJones, line 101: Can we mention that they should always call `restore_current_blog()` afterwards, since that is sometimes a cause of difficult to determine logic bugs? I would say that encouraging that, is more important than telling them how `switch_to_blog()` works.

rebeccahum: @GaryJones I wonder if it would be better to have a supplementary sniff to flag for a missing `restore_current_blog()` after a `switch_to_blog()` with that type of messaging recommendation?

GaryJones: Sure. I think there may even be one around somewhere.

jrfnl: I agree with @rebeccahum that the check for `restore_current_blog()` would probably best be done in a separate sniff (and would actually be a good one to add to WPCS). 👍

jrfnl (commented):
As per https://github.com/Automattic/VIP-Coding-Standards/issues/643#issuecomment-819803196, the change for this notification from `error`to `warning` can be safely added to this PR.

Other than that, looks good.

jrfnl (approved):

PR #668, opened 2021-04-20
ProperEscapingFunction: fix overreach

### LINKED ISSUE #667: False positive for notAttrEscAttr in ProperEscapingFunction sniff when printf()/sprintf() used
## Bug Description

When updating from 2.2.0 to 2.3.0 we got a lot of false positives of `notAttrEscAttr` in the `ProperEscapingFunction` sniff where `printf()` or `sprintf()` are being used. For example:

```php
printf( '<meta name="generator" content="%s">', esc_attr( $content ) );
```

This code is getting flagged with:

> Wrong escaping function, using `esc_attr()` in a context outside of HTML attributes may not escape properly.

See other such instances in the PR's check annotations: https://github.com/ampproject/amp-wp/pull/6079/files

## Minimal Code Snippet

See above.

## Error Code

`WordPressVIPMinimum.Security.ProperEscapingFunction.notAttrEscAttr`

## Environment

| PHP version             | 7.4.16 
| PHP_CodeSniffer version | 3.6.0
| VIPCS version           | 2.3.0

jrfnl:
As part of the changes made in #624, the `T_COMMA` token was added to the list of tokens to skip over, to allow for `echo` statements with multiple arguments passed as a coma-delimited list.

As a side-effect, this caused the sniff to also examine `[s]printf()`-like function calls where the first parameter is a text string, while the second is often a variable within a call to one of the escaping functions.

The current change fixes this by only adding the `T_COMMA` token to the "ignore when looking for the previous token"-list when in an `echo` statement.

Includes unit test.

Fixes #667

Additional notes:
* I've run the sniff over WP Core to verify the fix and have verified that all 23 violations being throw up are correctly detected violations.
* If it would be considered a good idea to also examine, `[s]printf()`-like function calls for this sniff for proper escaping, I suggest opening a separate, new feature request as that is change would need significantly different and quite complex logic and does not fall within the scope of this bug fix.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
@@ -46,7 +46,6 @@ class ProperEscapingFunctionSniff extends Sniff {
 		T_OPEN_TAG           => T_OPEN_TAG,
 		T_OPEN_TAG_WITH_ECHO => T_OPEN_TAG_WITH_ECHO,
 		T_STRING_CONCAT      => T_STRING_CONCAT,
-		T_COMMA              => T_COMMA,
 		T_NS_SEPARATOR       => T_NS_SEPARATOR,
 	];
 
@@ -107,7 +106,13 @@ class ProperEscapingFunctionSniff extends Sniff {
 			return;
 		}
 
-		$html = $this->phpcsFile->findPrevious( $this->echo_or_concat_tokens, $stackPtr - 1, null, true );
+		$ignore             = $this->echo_or_concat_tokens;
+		$start_of_statement = $this->phpcsFile->findStartOfStatement( $stackPtr, T_COMMA );
+		if ( $this->tokens[ $start_of_statement ]['code'] === T_ECHO ) {
+			$ignore[ T_COMMA ] = T_COMMA;
+		}
+
+		$html = $this->phpcsFile->findPrevious( $ignore, $stackPtr - 1, null, true );
 
 		// Use $textStringTokens b/c heredoc and nowdoc tokens will never be encountered in this context anyways..
 		if ( $html === false || isset( Tokens::$textStringTokens[ $this->tokens[ $html ]['code'] ] ) === false ) {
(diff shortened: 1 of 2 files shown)


westonruter:
It works much better but there are still false positives involving `esc_attr_x()`:

```
FILE: includes/validation/class-amp-validated-url-post-type.php
----------------------------------------------------------------------
FOUND 5 ERRORS AFFECTING 5 LINES
----------------------------------------------------------------------
 2370 | ERROR | Wrong escaping function, using `esc_attr_x()` in a
      |       | context outside of HTML attributes may not escape
      |       | properly.
      |       | (WordPressVIPMinimum.Security.ProperEscapingFunction.notAttrEscAttr)
 2379 | ERROR | Wrong escaping function, using `esc_attr_x()` in a
      |       | context outside of HTML attributes may not escape
      |       | properly.
      |       | (WordPressVIPMinimum.Security.ProperEscapingFunction.notAttrEscAttr)
 2425 | ERROR | Wrong escaping function, using `esc_attr_x()` in a
      |       | context outside of HTML attributes may not escape
      |       | properly.
      |       | (WordPressVIPMinimum.Security.ProperEscapingFunction.notAttrEscAttr)
 2456 | ERROR | Wrong escaping function, using `esc_attr_x()` in a
      |       | context outside of HTML attributes may not escape
      |       | properly.
      |       | (WordPressVIPMinimum.Security.ProperEscapingFunction.notAttrEscAttr)
 2461 | ERROR | Wrong escaping function, using `esc_attr_x()` in a
      |       | context outside of HTML attributes may not escape
      |       | properly.
      |       | (WordPressVIPMinimum.Security.ProperEscapingFunction.notAttrEscAttr)
----------------------------------------------------------------------
```

See [annotations in PR](https://github.com/ampproject/amp-wp/pull/6079/files#diff-6d0df4c4b1b89108880dd3988ff8d0c50b71b22a6a95e12db84373f9db6be217).

westonruter:
Sorry, I was mistaken. Our code was using `esc_attr_x()` when we should have been using `esc_html_x()`, so the sniff caught a big! True positive!

jrfnl:
Thanks for testing @westonruter and glad to hear that this PR fixes the issue.

rebeccahum (approved):
Tested and works great!

manooweb:
Hello,
When this fix will be released please?

rebeccahum:
@manooweb Hi! This will be released with 2.3.1, which we are hoping to within the next few days.

manooweb:
Ok 👍 thanks

Issue #554, opened 2020-07-27
Incorrect message for Security.ProperEscapingFunction.htmlAttrNotByEscHTML


GaryJones:
## Bug Description

The message for `WordPressVIPMinimum.Security.ProperEscapingFunction.htmlAttrNotByEscHTML` doesn't allow for the possibility of the value being a URL, and therefore `esc_url()` being correct instead of `esc_attr()` i.e. it gives the wrong advice in some circumstances.

## Minimal Code Snippet

```php
<form method="post" action="<?php echo esc_html(admin_url('admin.php?page='.$base_name.'&amp;mode=logs&amp;id='.$poll_id)); ?>">
```

That gave the following message:

> 🚫 Error: Wrong escaping function. HTML attributes should be escaped by `esc_attr()`, not by `esc_html()` (_WordPressVIPMinimum.Security.ProperEscapingFunction.htmlAttrNotByEscHTML_).

Since the value is a URL, the correct escaping would be `esc_url()`.

## Error Code

WordPressVIPMinimum.Security.ProperEscapingFunction.htmlAttrNotByEscHTML

## Environment

| Question                | Answer
| ------------------------| -------
| PHP version             | 7.?.? 
| PHP_CodeSniffer version | 3.5.5
| VIPCS version           | 2.1.0

## Additional Context (optional)

In the simplest fix, the message just needs an "or `esc_url()`" added to it.

We could potentially check if the attribute is an `src` or `href`, or `action` on certain elements, or if the call contained a URL-related function (`home_url()`, `admin_url()`, etc.) to give a tighter message, but there will always be cases of `data-*` attributes, or simple variables where we couldn't determine for certain if the value was meant to be a URL. The simple change in message avoids the completely incorrect message for some situations.

## Tested Against `master` branch?

- [x] I have verified the issue still exists in the `master` branch of VIPCS.
- [x] I have verified the issue still exists in the `develop` branch of VIPCS.

GaryJones:
Looking at https://github.com/Automattic/VIP-Coding-Standards/blob/develop/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php, there is already logic that looks for `href` or `src` attributes and gives a message regarding `esc_url()` already. It just needs to add support for `action` as well, and rename the method to allow for scaling to other attributes than just `href` and `src`.

PR #670, opened 2021-04-20
ProperEscapingFunction: improve "action" match precision

### LINKED ISSUE #669: False positive for ProperEscapingFunction.hrefSrcEscUrl when attribute with action in name is used
## Bug Description

The above sniff is extremely prone to false positives, as it just checks for action/src/href, but should be at least be more specific for action.

## Minimal Code Snippet

```php
<input data-action="<?php echo esc_attr( $my_var ); ?>">

or:
'https://demo.com?foo=bar&my-action='<?php echo esc_attr( $var ); ?>
```

## Error Code

Wrong escaping function. href, src, and action attributes should be escaped by `esc_url()`, not by `esc_attr()`.
WordPressVIPMinimum.Security.ProperEscapingFunction.hrefSrcEscUrl

## Environment

| PHP version             | 7.4.14
| PHP_CodeSniffer version | 3.6.0
| VIPCS version           | 2.3.0

jrfnl:
In VIPCS 2.2.0, checking for the use of `esc_url()` for "action" HTML attributes was introduced in PR #575 in response to issue #554.

As it is not inconceivable that "action" is used as a suffix for custom HTML attributes - like "data-action"- for which the value does not necessarily has to be a URL, the check for the "action" HTML attribute should make sure it is the complete attribute name and not used as a suffix for a custom attribute name.

This change contains a minor refactor of the code which examines the content of the previous text string.

Instead of looping over the various lists and doing a `substr()` on the same content 25 times, it will now use a regular expression to gather the necessary information to throw the right errors in one go.

Includes unit tests.

Fixes #669

Note: This PR removes two `private` properties which were introduced in #624/VIPCS 2.3.0 and two `public` methods.

The removal of the properties is safe. The removal of the `public` methods could be considered a BC-break as these methods were `public`, though they never should have been.

If so preferred, the `public` methods could be deprecated instead and remain in the code base as dead code/emptied out methods until the next major release upon which they could be removed.

---

### ProperEscapingFunction: fine-tune attribute regex

This adds test cases with:
* No space before "action" as it is at the start of the line in a multi-line text string.
* A tab before "action".

... and makes minor adjustments to the regex to safeguard handling these cases correctly.

### ProperEscapingFunction: deprecate, don't remove

... the `public` methods and as those use the `private` properties and extending sniffs may rely on the functionality of the `public` methods, we can't removed the `private` properties yet either, so deprecating those too.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
@@ -23,7 +23,7 @@ class ProperEscapingFunctionSniff extends Sniff {
 	 *
 	 * @var string
 	 */
-	const ATTR_END_REGEX = '`(?<attrname>href|src|url|\s+action)?=(?:(?:\\\\)?["\'])?$`i';
+	const ATTR_END_REGEX = '`(?<attrname>href|src|url|(^|\s+)action)?=(?:\\\\)?["\']*$`i';
 
 	/**
 	 * List of escaping functions which are being tested.
--- a/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php
@@ -57,6 +57,39 @@ class ProperEscapingFunctionSniff extends Sniff {
 	];
 
 	/**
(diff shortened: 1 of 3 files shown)


rebeccahum (dismissed):
Very clever! I love how you were able to eliminate the looping over and kill many birds with one stone.

rebeccahum (approved):

PR #675, opened 2021-04-22
ProperEscapingFunction: allow for comma's in first short open echo tag expression

### LINKED ISSUE #671: ProperEscapingFunction: account for T_OPEN_TAG_WITH_ECHO
## Bug Description

Introduced in #668, we should also account for `T_OPEN_TAG_WITH_ECHO`:

```php
<?= '<h2>' , esc_attr( $test ) , '</h2>'; // Warning.
```

However, it's not as simple as just adding it in an `OR` statement to:

https://github.com/Automattic/VIP-Coding-Standards/blob/7decb5b1f0570c69b66007394952d970643d073c/WordPressVIPMinimum/Sniffs/Security/ProperEscapingFunctionSniff.php#L111

This is because `findStartOfStatement()` will not recognize `<?=` as the start of the statement if there is inline HTML before it, e.g.

```php
<h1>TEST</h1>

<?= '<h2>' , esc_attr( $test ) , '</h2>'; // Warning.
```

However, using just the below will return the expected `$start_of_statement`:

```php
<?= '<h2>' , esc_attr( $test ) , '</h2>'; // Warning.
```

## Error Code

WordPressVIPMinimum.Security.ProperEscapingFunction.notAttrEscAttr

jrfnl:
Short open echo tags will act as an echo for the first expression and allow for passing multiple comma-separated parameters.
However, short open echo tags also allow for additional statements after, but those have to be full PHP statements, not expressions.

So, the `T_COMMA` token should be allowed and skipped over in the first expression, but not for subsequent statements following a short open echo tag.

`$phpcsFile->findStartOfStatement()` unfortunately is useless - even in its fixed-up form as will be in PHPCS 3.6.1-, as it will return the first token in the statement, which can be anything - variable, text string - without any indication of whether this is the start of a normal statement or a short open echo expression.
So, if we used that, we'd still need to walk back from every start of statement to the previous non-empty to see if it is the short open echo tag.

So to solve this conundrum, I've implemented a simple tracking system which will keep track of whether we have seen a short open echo tag and are within the first statement (expression) after this tag and will add the `T_COMMA` token if those conditions are fulfilled.

Fixes #671


jrfnl:
Rebased for merge conflicts (without changes other than to resolve the conflicts)

rebeccahum (approved):

PR #692, opened 2021-06-24
Remove dbDelta Warning

### LINKED ISSUE #479: Update output when dbDelta is found in the codebase
When a codebase is scanned and `dbDelta` is found the following error is output:

> All database modifications have to approved by the WordPress.com VIP team.

https://github.com/Automattic/VIP-Coding-Standards/blob/master/WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php#L94

While this was previously true on WordPress.com we no longer require the VIP team to approve all database changes at every code review level. 

As this can still cause performance problems, we should continue to flag this as a possible issue. However, instead of stating that the VIP team needs to approve the change, we could link to documentation of some things to check or other helpful tips.

BrookeDot:
This removes the dbDelta message which previously stated to contact
WordPress.com VIP support for review. Fixes #479.


GaryJones:
There are various other references to `dbDelta()` which will need updating for the checks to pass.

https://github.com/Automattic/VIP-Coding-Standards/search?q=dbDelta

rebeccahum (approved):

PR #693, opened 2021-06-30
Downgrade htmlAttrNotByEscHTML to a warning

### LINKED ISSUE #601: Convert htmlAttrNotByEscHTML to warning or back to error for consistency with other rules
We typically don't have any errors that are below 5 (with 5 being the default severity level).

https://github.com/Automattic/VIP-Coding-Standards/blob/60ad148168258afccd05ff0429d53a7e810270a5/WordPress-VIP-Go/ruleset.xml#L259-L261

However, on the VIP Go ruleset, we have `WordPressVIPMinimum.Security.ProperEscapingFunction.htmlAttrNotByEscHTML` marked as an error at a level 3. I think this could cause potential confusion and for consistency's sake, we should either mark it as a warning with a higher severity or bring it back to the default error level. 

FWIW, I don't have the exact context on why it was brought down to a level 3 and [the PR where it was introduced](https://github.com/Automattic/VIP-Coding-Standards/commit/b8c5b767e96d32bbfa096a2d9e07f7db382737d9) has no description.

rebeccahum:
Fixes #601.


jrfnl (approved):
I haven't tested, but LGTM.

PR #706, opened 2021-12-29
README: update for Composer 2.2

jrfnl:
The `dealerdirect/phpcodesniffer-composer-installer` Composer plugin is a non-dev requirement for VIPCS.

As of Composer 2.2.0, Composer plugins need to be explicitly allowed to run.

This commit adds the CLI command to set those permissions to the installation instructions.

Refs:
* https://blog.packagist.com/composer-2-2/#more-secure-plugin-execution

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,23 @@ Go to https://docs.wpvip.com/technical-references/code-review/phpcs-report/ to l
 
 ## Installation
 
-`composer require automattic/vipwpcs`, or `composer g require automattic/vipwpcs` if installing globally. 
+To install the VIP Coding Standards, run:
+```bash
+composer config allow-plugins.dealerdirect/phpcodesniffer-composer-installer true
+composer require automattic/vipwpcs
+```
+
+or if installing globally:
+```bash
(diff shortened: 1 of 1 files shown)


GaryJones:
Is this still needed if the composer.json has got the updated configuration?

jrfnl:
> Is this still needed if the composer.json has got the updated configuration?

Yes, every project needs to give permission for plugins to run on their project. It's not as if the permission given in this project propagates to consumer projects (that would negate the security measure side of things as that way a dependency could still introduce the running of a malicious plugin without a project realizing).

Just to be sure, I did a test run anyway, with and without the `config` setting in this package, i.e. `require automattic/vipwpcs:"2.3.3"` (without) and `require automattic/vipwpcs:"dev-master"` (with) and yes, I got the prompt to grant permission in both cases.

GaryJones (approved):

rebeccahum:
@GaryJones Do we need to update our public docs for [installation instructions](https://docs.wpvip.com/how-tos/php_codesniffer/) as well?

PR #721, opened 2022-08-24
Update composer.json keywords

GaryJones:
As per https://getcomposer.org/doc/04-schema.md#keywords by including "static analysis", then later versions of Composer will prompt users if the package is installed with `composer require` instead of `composer require-dev`.


jrfnl (approved):
👍🏻

rebeccahum (approved):

Issue #727, opened 2022-10-24
Make VIPCS PHP 8-friendlier


rebeccahum:
## Bug Description

A lot of users are running into https://github.com/WordPress/WordPress-Coding-Standards/pull/1935 since we're waiting on WPCS 3.0 and the current version of 2.3.0 does not have the fix included. Creating new issue for tracking purposes.

swissspidy:
Would it be an option to change the `"wp-coding-standards/wpcs": "^2.3"` requirement to `"wp-coding-standards/wpcs": "^2.3 || dev-develop"?

This way, consumers with `minimum-stability: 'dev'` could use the develop branch including the fix, while everyone else can still be on 2.3.0. Assuming of course that this package here works well together with the develop version.

Haven't tested this though, but I think it should work.

christianc1:
Could we change the requirement to `"wp-coding-standards/wpcs": "^2.3 || dev-master#2.3.0-fix-vsprintf"`?  Opened #730 if this is viable.

jrfnl:
FYI: we're getting close to releasing WPCS 3.0.0. No promises on timeline, but I'm hoping within the next month.

ThirdEyeRose:
@jrfnl , I assume the timeline for the release of WPCS 3.0.0 has shifted? Do you know if there's an updated ETA for WPCS 3.0.0?

PR #730, opened 2022-11-22
Open wpcs composer dependency to allow vsprint fixed tag

christianc1:
Related to #727 , I think this change should allow a project to use https://github.com/WordPress/WordPress-Coding-Standards/releases/tag/2.3.0-fix-vsprintf without also pulling the unrelated (and possibly breaking) changes from wpcs@develop.


jrfnl:
The [2.3.0-fix-vsprintf](https://github.com/WordPress/WordPress-Coding-Standards/releases/tag/2.3.0-fix-vsprintf) tag _already_ includes breaking changes, so I don't think this is the best of ideas.

christianc1:
I see.  Oh well, appreciate the context @jrfnl.

PR #732, opened 2022-12-20
WPCS 3.0: Composer updates

GaryJones:
We need VIPCS to be compatible with WPCS 3.0, so this is the first of a few PRs to address the known compatibility issues. 

### Composer: up the minimum PHPCS version to 3.7.1

PHPCS 3.7.1 adds support for PHP 8.1 features, including fixing a retokenization of reserved keywords bug in PHPCS 3.7.0. Amends a couple of GitHub workflows for this change too.

### Composer: rename scripts

Brings them more in line with the Composer scripts in WPCS.

### Composer: add script descriptions

Displayed for `composer list`.

### Composer: use develop branch of WPCS

VIPCS must be compatible with WordPress Coding Standards 3.0.0, currently in development on its `develop` branch. WPCS now uses some extra dependencies, which are also in development, so we need to allow for a `minimum-stability` of `dev`. Run `composer update -W` to pull those extra dependencies in.

### Composer: Normalize the order

Uses the composer-normalize plugin to keep a consistent order.

**Note:** GitHub Actions may report failures whilst this and the next few PRs go through, but I don't want to bundle all the changes into big PRs.

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -30,15 +30,11 @@ jobs:
             phpcs_version: 'dev-master'
             wpcs_version: '2.3.*'
           - php: '5.4'
-            phpcs_version: '3.5.5'
+            phpcs_version: '3.7.1'
             wpcs_version: '2.3.*'
 
           - php: 'latest'
             phpcs_version: 'dev-master'
-            wpcs_version: '2.3.*'
-          - php: 'latest'
-            # PHPCS 3.6.1 is the lowest version of PHPCS which supports PHP 8.1.
-            phpcs_version: '3.6.1'
             wpcs_version: '2.3.*'
 
     name: "QTest${{ matrix.phpcs_version == 'dev-master' && ' + Lint' || '' }}: PHP ${{ matrix.php }} - PHPCS ${{ matrix.phpcs_version }}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,7 +71,7 @@ jobs:
       # - experimental: Whether the build is "allowed to fail".
       matrix:
         php: ['5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4']
-        phpcs_version: ['3.5.5', 'dev-master']
+        phpcs_version: ['3.7.1', 'dev-master']
         wpcs_version: ['2.3.*']
         experimental: [false]
 
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
 	},
 	"scripts-descriptions": {
 		"check-all": "Run all checks (lint, phpcs, feature completeness) and tests.",
-		"check-complete": "Check if all the sniffs have tests.",
+		"check-complete": "Check if all the sniffs have unit tests.",
 		"check-complete-strict": "Check if all the sniffs have unit tests and XML documentation.",
 		"check-cs": "Run the PHPCS script against the entire codebase.",
 		"lint": "Lint PHP files against parse errors.",
--- a/composer.json
+++ b/composer.json
@@ -35,25 +35,27 @@
 		}
 	},
 	"scripts": {
-		"install-codestandards": "Dealerdirect\\Composer\\Plugin\\Installers\\PHPCodeSniffer\\Plugin::run",
-		"ruleset": "bin/ruleset-tests",
+		"check-all": [
+			"@lint",
+			"@check-cs",
+			"@run-tests",
+			"@check-complete-strict",
+			"@ruleset"
+		],
+		"check-complete": [
+			"@php ./vendor/phpcsstandards/phpcsdevtools/bin/phpcs-check-feature-completeness -q ./WordPressVIPMinimum"
+		],
+		"check-complete-strict": [
+			"@php ./vendor/phpcsstandards/phpcsdevtools/bin/phpcs-check-feature-completeness ./WordPressVIPMinimum"
+		],
+		"check-cs": "bin/phpcs",
 		"lint": [
 			"bin/php-lint",
 			"bin/xml-lint"
 		],
-		"phpcs": "bin/phpcs",
-		"phpunit": "bin/unit-tests",
-		"coverage": "bin/unit-tests-coverage",
-		"check-complete": [
-			"@php ./vendor/phpcsstandards/phpcsdevtools/bin/phpcs-check-feature-completeness -q ./WordPressVIPMinimum"
-		],
-		"test": [
-			"@lint",
-			"@ruleset",
-			"@phpunit",
-			"@phpcs",
-			"@check-complete"
-		]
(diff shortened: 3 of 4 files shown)


.github/workflows/quicktest.yml
```diff
@@ -30,16 +30,12 @@ jobs:
...
+            phpcs_version: '3.7.1'
             wpcs_version: '2.3.*'
 
           - php: 'latest'
             phpcs_version: 'dev-master'
             wpcs_version: '2.3.*'
-          - php: 'latest'
-            # PHPCS 3.6.1 is the lowest version of PHPCS which supports PHP 8.1.
-            phpcs_version: '3.6.1'
-            wpcs_version: '2.3.*'
```
jrfnl, line 42: Why are you removing this build ?

The "Quick test" workflow was set up to run the tests against the lowest and highest supported PHP/PHPCS combinations. The build you are removing is the "high PHP/low PHPCS".

I suggest bringing it back, removing the comment and changing the `phpcs_version` to `3.7.1` instead.

.github/workflows/test.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       # - experimental: Whether the build is "allowed to fail".
       matrix:
         php: ['5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4']
-        phpcs_version: ['3.5.5', 'dev-master']
+        phpcs_version: ['3.7.1', 'dev-master']
```
jrfnl, line 74: The comment on line 67 can also be removed.

composer.json
```diff
@@ -1,63 +1,75 @@
...
+		"run-tests": "bin/unit-tests",
+		"run-tests-coverage": "bin/unit-tests-coverage"
 	},
-	"support": {
-		"issues": "https://github.com/Automattic/VIP-Coding-Standards/issues",
-		"wiki": "https://github.com/Automattic/VIP-Coding-Standards/wiki",
-		"source": "https://github.com/Automattic/VIP-Coding-Standards"
+	"scripts-descriptions": {
+		"check-all": "Run all checks (lint, phpcs, feature completeness) and tests.",
+		"check-complete": "Check if all the sniffs have tests.",
```
jrfnl, line 69:
```suggestion
		"check-complete": "Check if all the sniffs have unit tests.",
```

The script doesn't check for "ruleset tests" which this repo also has, so let's make that clear.

composer.json
```diff
@@ -1,63 +1,75 @@
...
+		"check-complete-strict": "Check if all the sniffs have unit tests and XML documentation.",
+		"check-cs": "Run the PHPCS script against the entire codebase.",
```
jrfnl, line 71:
```suggestion
		"check-cs": "Check the code style and code quality of the codebase via PHPCS.",
```

I believe it is more descriptive to describe what the script checks than what script is being run...

Also, PHPCS CLI options like `--sniffs=...` or passing a specific path (should) still work, even when running the CS check via the script.

[Edit] Note: just looked at the script in `bin` and it looks like it doesn't support passing CLI args, so I guess that `bin` script should be improved (or just removed and replaced by having `@php ./vendor/squizlabs/php_codesniffer/bin/phpcs` directly in the Composer script.

composer.json
```diff
@@ -1,63 +1,75 @@
...
+		"lint": "Lint PHP files against parse errors.",
```
jrfnl, line 72:
```suggestion
		"lint": "Check PHP files against parse errors.",
```

"Checking against parse errors" = "linting"

Also, looking at the script, the `lint` command also runs `xmllint`, should that be mentioned ?

composer.json
```diff
@@ -1,63 +1,75 @@
...
+		"run-tests": "Run all the unit tests for the VIP Coding Standards sniffs."
```
jrfnl, line 73:
```suggestion
		"run-tests": "Run the unit tests for the VIP Coding Standards sniffs."
```

People can still use PHPUnit options like `--filter ...`, even when running the tests via the script.

composer.json
```diff
@@ -1,63 +1,75 @@
...
+		],
+		"ruleset": "bin/ruleset-tests",
+		"run-tests": "bin/unit-tests",
+		"run-tests-coverage": "bin/unit-tests-coverage"
 	},
-	"support": {
-		"issues": "https://github.com/Automattic/VIP-Coding-Standards/issues",
-		"wiki": "https://github.com/Automattic/VIP-Coding-Standards/wiki",
-		"source": "https://github.com/Automattic/VIP-Coding-Standards"
+	"scripts-descriptions": {
```
jrfnl, line 67: The description for the `ruleset` and the `run-tests-coverage` scripts are missing.

composer.json
```diff
@@ -1,63 +1,75 @@
...
 	},
+	"minimum-stability": "dev",
+	"prefer-stable": true,
 	"config": {
 		"allow-plugins": {
 			"dealerdirect/phpcodesniffer-composer-installer": true
 		}
 	},
 	"scripts": {
-		"install-codestandards": "Dealerdirect\\Composer\\Plugin\\Installers\\PHPCodeSniffer\\Plugin::run",
```
jrfnl, line 38: Just mentioning for the record: this commit also removed the `install-codestandards` script, which should no longer be needed. ✅

jrfnl (changes requested):
Hi @GaryJones Thanks for setting this up. The PR is mixing a few changes with different decision points in one PR (Composer normalization and scripts changes vs changing versions of dependencies), but that's no biggie.

I've left a number of comments inline for your perusal.

Other than that, I'd like to suggest _not_ changing the WPCS version to `develop` as that's a moving target.

I'm still hoping we reach the point where we can tag a `3.0.0-RC` release before I leave for my break. Once that's tagged, the WPCS version for this repo could then be set to `3.0.0` (with `minimum-stability` set to `RC`), which makes it less of a moving target + makes it more descriptive what's going on.

jrfnl:
Note: if the change to WPCS `develop` would stay, the CI workflows should be updated for that change in the same commit.

composer.json
GaryJones: Handled in e0ea18949f0cfc532fc5a0e8289144fa744b34b4.

composer.json
GaryJones: Handled in e0ea18949f0cfc532fc5a0e8289144fa744b34b4.

.github/workflows/test.yml
GaryJones: I like having the contextual information handy, so I've updated the doc comment to make more sense, in b58668ab808fbdc46ad55aac58b78021ac8863b0.

composer.json
GaryJones: Handled in 0c63b7ec751cbadf0a84d518884d3ae44abed505.

GaryJones:
The changes here are mostly now handled in
- #741
- #742

I've skipped the normalizing of the `composer.json` for now.

PR #733, opened 2022-12-20
WPCS 3.0: Fix coding standards of VIPCS sniffs

GaryJones:
~**Once #732 has been merged, this can be rebased** to just show coding standards changes.~

----

Update the VIPCS sniffs to fix the coding standards changes with WPCS 3.0.

### CS: Use pre-increment over post-increment

Slightly better performance.

### CS: Avoid reserved keyword for function param name

Using named parameters with a variable named `$string` is unfavourable.

### ~CS: Update WordPress-Extra exclusions~

~WordPress-Extra will no longer use the `Generic.Arrays.DisallowShortArraySyntax` rule, but does use the improved `Universal.Arrays.DisallowShortArraySyntax` from PHPCSExtra instead, so whilst we have short array syntax being used in the VIPCS sniffs, let's continue to exclude the rule.~

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Classes/DeclarationCompatibilitySniff.php
+++ b/WordPressVIPMinimum/Sniffs/Classes/DeclarationCompatibilitySniff.php
@@ -291,7 +291,7 @@ class DeclarationCompatibilitySniff extends AbstractScopeSniff {
 					return;
 				}
 			}
-			$i++;
+			++$i;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


jrfnl (commented):
@GaryJones Reviewed just the commits which belong to this PR.

IMO...:
* Commit 1 and 2 (increment + reserved keywords) can/should be rebased on `develop` and pulled separately. These commits can be merged independently of PR #732 as they just apply some best practices (and anticipate that those will be checked for in WPCS 3.x).
* By having those in a separate PR and merging it _ahead_ of PR #732, PR #732 has a higher chance of getting a passing build.
* Commit 3 (ruleset change) should be combined with the commit changing the minimum WPCS version to `3.x` in Composer as as soon as `3.x` becomes the minimum, the ruleset becomes invalid without the change from this commit.

GaryJones:
Updated to just the two suggested CS-related commits. No longer requires #732.

jrfnl (approved):

jrfnl:
As this has been merged into `develop`, I've moved this to the `2.3.4` milestone.

PR #734, opened 2022-12-20
WPCS 3.0: Fix incompatibilities by using PHPCSUtils

GaryJones:
~Once #742 has been merged, this can be rebased to allow the workflow tests to run. #742 makes PHPCS 3.7.1 as the minimum version for VIPCS, which is the constraint on PHPCSUtils 1.0.~ Now rebased.

----

Several methods were removed or moved from the Sniff class in WPCS, so these changes were spotted when running unit tests and getting errors. There are undoubtedly many more opportunities to use PHPCSUtils versions of functions to replace existing VIPCS code to improve the quality of the sniffs, but these are not tackled here.

These changes don't **require** WPCS 3.0 - we're just making the same types of changes that WPCS 3.0, by using PHPCSUtils as a direct dependency.

### Composer: Add PHPCSUtils as required dependency
VIPCS uses PHPCSUtils directly (such as for MessageHelper), so it should be marked as such in the Composer config.

### Sniff::strip_quotes(): switch to use the PHPCSUtils version
The original `Sniff::strip_quotes()` method was removed for WPCS 3.0.

### Sniff::find_array_open_close(): switch to use the PHPCSUtils version
The original `Sniff::find_array_open_close()` method was removed for WPCS 3.0.

### Sniff::get_function_call_parameter(): switch to use the PHPCSUtils version
The original `Sniff::get_function_call_parameter()` method was removed for WPCS 3.0.

### Use PHPCSUtils MessageHelper::addMessage
This is more tested than the WPCS `Sniff::addMessage()`, and the original was removed.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
@@ -9,7 +9,7 @@
 
 namespace WordPressVIPMinimum\Sniffs;
 
-use WordPressVIPMinimum\Sniffs\Sniff;
+use PHPCSUtils\Utils\MessageHelper;
 
 /**
  * Restricts usage of some variables.
@@ -200,11 +200,13 @@ abstract class AbstractVariableRestrictionsSniff extends Sniff {
 				continue;
 			}
 
-			$this->addMessage(
+			$code = MessageHelper::stringToErrorcode( $groupName . '_' . $match[1] );
+			MessageHelper::addMessage(
+				$this->phpcsFile,
 				$group['message'],
 				$stackPtr,
 				$group['type'] === 'error',
-				$this->string_to_errorcode( $groupName . '_' . $match[1] ),
+				$code,
 				[ $var ]
 			);
 
--- a/WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php
@@ -8,8 +8,9 @@
 
 namespace WordPressVIPMinimum\Sniffs\Constants;
 
-use WordPressVIPMinimum\Sniffs\Sniff;
 use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Utils\PassedParameters;
+use WordPressVIPMinimum\Sniffs\Sniff;
 
 /**
  * Sniff for properly using constant name when checking whether a constant is defined.
@@ -55,7 +56,7 @@ class ConstantStringSniff extends Sniff {
 			return;
 		}
 
-		$param = $this->get_function_call_parameter( $stackPtr, 1 );
+		$param = PassedParameters::getParameter( $this->phpcsFile, $stackPtr, 1, 'constant_name' );
 		if ( $param === false ) {
 			// Target parameter not found.
 			return;
--- a/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/DynamicCallsSniff.php
@@ -8,6 +8,7 @@
 namespace WordPressVIPMinimum\Sniffs\Functions;
 
 use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Utils\TextStrings;
 use WordPressVIPMinimum\Sniffs\Sniff;
 
 /**
@@ -139,7 +140,7 @@ class DynamicCallsSniff extends Sniff {
 		 * If we reached the end of the loop and the $value_ptr was set, we know for sure
 		 * this was a plain text string variable assignment.
 		 */
-		$current_var_value = $this->strip_quotes( $this->tokens[ $value_ptr ]['content'] );
+		$current_var_value = TextStrings::stripQuotes( $this->tokens[ $value_ptr ]['content'] );
 
 		if ( isset( $this->disallowed_functions[ $current_var_value ] ) === false ) {
 			// Text string is not one of the ones we're looking for.
--- a/WordPressVIPMinimum/Sniffs/Hooks/AlwaysReturnInFilterSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Hooks/AlwaysReturnInFilterSniff.php
@@ -7,8 +7,9 @@
 
 namespace WordPressVIPMinimum\Sniffs\Hooks;
 
-use WordPressVIPMinimum\Sniffs\Sniff;
 use PHP_CodeSniffer\Util\Tokens;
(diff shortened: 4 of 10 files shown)


GaryJones:
Tests pass fine locally, so not sure why GitHub Actions is having an issue finding and using PHPCSUtils.

WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php
```diff
@@ -55,7 +56,7 @@ public function process_token( $stackPtr ) {
 			return;
 		}
 
-		$param = $this->get_function_call_parameter( $stackPtr, 1 );
+		$param = PassedParameters::getParameter( $this->phpcsFile, $stackPtr, 1 );
```
jrfnl, line 59:
```suggestion
		$param = PassedParameters::getParameter( $this->phpcsFile, $stackPtr, 1, 'constant_name' );
```

For function calls, the parameter name is (soft) required now to allow for support for PHP 8.0+ function calls passing named parameters.

If the parameter name is not passed and the parameter being requested is not found, the `PassedParameters::getParameter()` method will throw an exception.

In this case, the parameter name for both functions being "sniffed" is the same, so can just be passed like per the suggestion, though if the sniff would at some point need to examine other functions, a different, more flexible, solution would be needed.

Refs:
* https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Utils-PassedParameters.html#method_getParameter
* https://www.php.net/manual/en/function.define.php
* https://www.php.net/manual/en/function.defined.php

WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
```diff
@@ -9,7 +9,7 @@
 
 namespace WordPressVIPMinimum\Sniffs;
 
-use WordPressVIPMinimum\Sniffs\Sniff;
```
jrfnl, line 12: The `WordPressVIPMinimum\Sniffs\Sniff` class is still being used....

WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
```diff
@@ -9,7 +9,7 @@
...
+use PHPCSUtils\Utils\MessageHelper;
```
jrfnl, line 12: This change is in the wrong commit (in "Add Sniff::merge_custom_array()", while it belongs with "Use PHPCSUtils MessageHelper::addMessage").

WordPressVIPMinimum/Sniffs/Sniff.php
```diff
@@ -17,4 +17,45 @@
  * @package VIPCS\WordPressVIPMinimum
  */
 abstract class Sniff extends \WordPressCS\WordPress\Sniff {
+	/**
```
jrfnl, line 20: Yes, the method in WPCS is now marked as not part of the "public API", but it can still be _used_.

The "not part of the public API", just means "use at your own risk as there may be BC-breaking changes made in non-major releases", but until such a time that such a BC-breaking change would be made, VIPCS could still use it as the functionality hasn't changed (at this time and I don't currently have any plans to change it either).

So based on that, I have two questions:
1. Why copy this method ?
2. If copying is really the way to go, why not follow the example set by WPCS and have it in a separate `Helper` class instead of adding it to the `Sniff` class ?
    Having it in a separate `Helper` class, will allow for the method to still be available to sniffs if any of them switch over to `abstract` base classes from PHPCSUtils in the future (which is exactly the reason why it was moved to a `Helper` in WPCS).

.github/workflows/unit-tests.yml
```diff
@@ -70,41 +70,19 @@ jobs:
       #   no additional versions are included in the array.
       # - experimental: Whether the build is "allowed to fail".
       matrix:
```
jrfnl, line 72: What with PHPCSUtils now being added as a dependency, should it be added to the matrix ? Or maybe the matrix/workflow should be changed to make use of the Composer `--prefer-lowest`/`--prefer-highest` options to simplify the maintenance of the version combinations ?

(yes, something I'm looking to do for WPCS as well)

jrfnl (commented):
Reviewed with the exception that I haven't done a detailed review of the last two commits (see below).

Along the same lines as my comment on #733:
* Most of this PR does not depend on PR #732. These changes can be made independently of the WPCS update (and in anticipation thereof).
* The first commit should be adding the PHPCSUtils package as a dependency as the other commits all depend on that package being available (the "Composer: Add PHPCSUtils as required dependency" commit).
* If commit 4 ("Add Sniff::merge_custom_array()") stays, I'd suggest squashing commit 6 ("CS: fix coding standards for merge_custom_array()") into that same commit.
* The last two commits ("CI: Refresh unit tests workflow" and "Disable ruleset test") do not belong with this PR.
    _I actually had a PR ready for most of those changes too, but got stuck on debugging why the ruleset test wasn't working, which is why I hadn't pulled it yet._

I'd recommend rebasing the PR (without those last two commits) on `develop` and changing the commit order.

WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
GaryJones: It is, but the `namespace` of this Abstract is the same as `Sniff`, so the `use` statement is not needed.

.github/workflows/unit-tests.yml
GaryJones: https://github.com/Automattic/VIP-Coding-Standards/issues/743 opened to handle this separately.

WordPressVIPMinimum/Sniffs/Sniff.php
GaryJones: Opened #744 to handle separately because all of the rest of this PR can then be framed as "Make some use of PHPCSUtils", rather than "Be compatible with WPCS 3.0".

WordPressVIPMinimum/Sniffs/Constants/ConstantStringSniff.php
GaryJones: Updated in 98f52dbd65c31d7574028a8d68419f0232ca9e68.

WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
GaryJones: The "Add Sniff::merge_custom_array()" commit was removed (to be handled by #744), but this line 12 change was added to db681ccf2d61dcac745c510c27ceff532f91b6db.

GaryJones:
> I'd recommend rebasing the PR (without those last two commits) on develop and changing the commit order.

Done.

PR #735, opened 2022-12-20
Unit tests: Support PHP >= 8.1

GaryJones:
While the different logic for PHP 8.1 was in the GitHub workflow, moving it to the bin script means the GitHub workflow can be simplified, and the same Composer script can be run locally too.

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -96,7 +96,6 @@ jobs:
         run: ./bin/php-lint
 
       - name: Run the unit tests
-        if: matrix.php != 'latest'
         run: ./bin/unit-tests
 
       - name: Run the ruleset tests
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -95,13 +95,9 @@ jobs:
         if: matrix.phpcs_version == 'dev-master'
         run: ./bin/php-lint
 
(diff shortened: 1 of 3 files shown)


jrfnl (commented):
Technically this change looks correct, but I can't test it as these scripts don't work on Windows anyway, so 🤷🏻‍♀️

.github/workflows/quicktest.yml
```diff
@@ -95,13 +95,9 @@ jobs:
         if: matrix.phpcs_version == 'dev-master'
         run: ./bin/php-lint
 
-      - name: Run the unit tests - PHP 5.4 - 8.0
+      - name: Run the unit tests
         if: matrix.php != 'latest'
```
GaryJones, line 99: @rebeccahum Did you mean to remove this as well?

.github/workflows/test.yml
```diff
@@ -156,13 +156,9 @@ jobs:
           composer-options: --ignore-platform-reqs
           custom-cache-suffix: $(date -u -d "-0 month -$(($(date +%d)-1)) days" "+%F")
 
-      - name: Run the unit tests - PHP 5.4 - 8.0
+      - name: Run the unit tests
         if: matrix.php < '8.1'
```
GaryJones, line 160: @rebeccahum Did you mean to remove this as well?

rebeccahum (approved):

PR #736, opened 2022-12-20
WPCS 3.0: Fix ruleset tests

GaryJones:
Once #734 has been merged, this can be rebased to just show the ruleset tests changes.

----

### Ruleset tests: update for WPCS 3.0

The ruleset tests were failing mainly due to the [renaming of some violation codes](https://github.com/WordPress/WordPress-Coding-Standards/pull/2108) that we were excluding.

The other change to account for was the [replacement of a WPCS StrictComparisons sniff with a PHPCSExtras Universal sniff](https://github.com/WordPress/WordPress-Coding-Standards/pull/1919). The sniff in PHPCSExtra contains a fixer. As this is a risky fixer, this fixer is turned off for WPCS. The sniff in PHPCSExtra will provide metrics about loose versus strict comparisons.

### Ruleset tests: Add label before test runs

Make it clear which ruleset being tested, even on a test failure.

### Enable ruleset tests

Now the issues have been address, these can be enabled again in `composer check-all` and a Github workflow.

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -28,14 +28,14 @@ jobs:
         include:
           - php: '5.4'
             phpcs_version: 'dev-master'
-            wpcs_version: '2.3.*'
+            wpcs_version: 'dev-develop'
           - php: '5.4'
             phpcs_version: '3.7.1'
-            wpcs_version: '2.3.*'
+            wpcs_version: 'dev-develop'
 
           - php: 'latest'
             phpcs_version: 'dev-master'
-            wpcs_version: '2.3.*'
+            wpcs_version: 'dev-develop'
 
     name: "QTest${{ matrix.phpcs_version == 'dev-master' && ' + Lint' || '' }}: PHP ${{ matrix.php }} - PHPCS ${{ matrix.phpcs_version }}"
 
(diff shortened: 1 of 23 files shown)


WordPress-VIP-Go/ruleset-test.inc
```diff
@@ -546,7 +546,7 @@ str_replace( 'foo', array( 'bar', 'foo' ), 'foobar' ); // Error.
 
 // WordPressVIPMinimum.Security.Underscorejs
 echo "<script>
- _.templateSettings = { 
+ _.templateSettings = {
```
jrfnl, line 549: Was this an _intentional_ or _accidental_ trailing whitespace ? I.e. are you changeing what's being tested now or not ?

(same question for the same change in `WordPressVIPMinimum/ruleset-test.inc`)

tests/RulesetTest.php
```diff
@@ -98,6 +98,9 @@ public function __construct( $ruleset, $expected = [] ) {
 			$this->phpcs_bin = realpath( $phpcs_bin );
 		}
 
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
jrfnl, line 101: There are quite a few ignore annotations of this sniff and the `system_calls` group of the `WordPress.PHP.DiscouragedPHPFunctions` sniff for this file.

Would it make sense to add an exception to the project `phpcs.xml.dist` file for those two rules for this particular file (and to remove the inline ignore annotations) ?
Alternatively, `// phpcs:disable` comments could be added to the top of this particular file for those error codes.

.github/workflows/quicktest.yml
```diff
@@ -91,6 +86,9 @@ jobs:
           composer-options: --ignore-platform-reqs
           custom-cache-suffix: $(date -u -d "-0 month -$(($(date +%d)-1)) days" "+%F")
 
+      - name: Display PHPCS installed standards
```
jrfnl, line 89: This doesn't validate anything. It just shows the standards, it doesn't error out if expected standards would not be registered.

Also, this information is already available at the bottom of the "Install Composer dependencies" step as the Composer PHPCS plugin will display it as part of its user facing output.

jrfnl (changes requested):
Reviewed with the exception that I haven't done a detailed review of the last commit (see below).

Along the same lines as previous reviews:
* Commit 1 ("Ruleset tests: update for WPCS 3.0") should be in the same PR as the commit changing the minimum WPCS version to 3.x in Composer as as soon as 3.x becomes the minimum, the rulesets becomes invalid without the changes from this commit.
* Commit 2 + 4 ("Ruleset tests: Add label before test runs" and "Ruleset tests: Update label") should be squashed together.
* Commit 3 shouldn't exist (nor should the "Disable ruleset test" commit in PR #734)
* Commit 5 ("Add PHPCSExtras dependency") doesn't belong with this PR and it is unclear why you are adding PHCPSExtra.
    For those sniffs in use by WPCS, the dependency is inherited from WPCS.
    If VIPCS is not directly using PHPCSExtra (adding sniffs from PHPCSExtra which are not used by WPCS), the dependency is not "our" dependency.
* Commit 6 ("CI: Use WPCS develop branch for quick test") should be combined with the "CI: Refresh unit tests workflow" commit from PR #734 and added to PR #732 as those changes are directly related to the change in the WPCS dependency version.

GaryJones:
Closing for now, as #746 handled the fixing of ruleset tests (and other bits) for PHP 8.1 while continuing to use WPCS 2.3.0.

Some of the changes (with feedback) in this PR will be pulled into a branch that uses WPCS 3.0.

PR #737, opened 2022-12-21
AdminBarRemovalUnitTest: Ensure final reset is read

GaryJones:
The `// phpcs:set` comment should be a PHP comment, but the PHP opening delimiter was missing.


jrfnl (approved):
Makes sense to me.

Note: it is more common for the `phpcs:set` directives to use `//`-style than `/* */`-style, though IIRC, both should work.

Also, `phpcs:set` can be used in "inline HTML", just don't add comment markers in that case.

PR #738, opened 2023-01-05
Composer: allow for the 1.0.0 version of the Composer PHPCS plugin

jrfnl:
The Composer PHPCS plugin has released its 1.0.0 version. :tada:

Important:
I've _widened_ the version constraints for the plugin, instead of _bumping_ it.

The reason for this is to prevent conflicts with end-user projects/other external PHPCS standards which may also require(-dev) the plugin, but may not (yet) have updated _their_ constraints for the plugin. If the version would have been bumped instead of widened, those users would get an unsolvable conflict during the `composer install` run (unless they `require-dev` the plugin for the root project, but then, that's exactly what we _don't_ want them to do as external standards managing the versions of the plugins should  be more reliable).

The version constraints now set cover all released versions which support external standards properly.

Ref: https://github.com/PHPCSStandards/composer-installer/releases/tag/v1.0.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
 	],
 	"require": {
 		"php": ">=5.4",
-		"dealerdirect/phpcodesniffer-composer-installer": "^0.4.1 || ^0.5 || ^0.6.2 || ^0.7",
+		"dealerdirect/phpcodesniffer-composer-installer": "^0.4.1 || ^0.5 || ^0.6.2 || ^0.7 || ^1.0",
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
✅

kasparsd:
Thanks for the fix! 

Is there a plan to release a stable version with these changes? It appears that the current diff is relatively small https://github.com/Automattic/VIP-Coding-Standards/compare/2.3.3..22733aa560f0f8855b848475770e4f800157e7f4 and contains mostly dev related changes.

PR #742, opened 2023-01-24
Composer: up the minimum PHPCS version to 3.7.1

GaryJones:
Add support for PHP 8.1 features, including fixing a retokenization of reserved keywords bug in PHPCS 3.7.0.

With a new minimum version of PHPCS, this allows consolidation and simplification of the testing matrix in the `test.yml` workflow.

I'd like to get this into 2.3.4, but this may be semantically better for 3.0.0?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,7 @@ on:
     paths-ignore:
       - '**.md'
   pull_request:
+
   # Allow manually triggering the workflow.
   workflow_dispatch:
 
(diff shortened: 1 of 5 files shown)


rebeccahum (approved):

GaryJones:
@rebeccahum Any preference on whether this could go into a 2.3.4, or hold on to a 3.0?

.github/workflows/test.yml
```diff
@@ -64,46 +65,18 @@ jobs:
...
-            phpcs_version: '3.6.1'
-            wpcs_version: '2.3.*'
-            experimental: false
-
-          # Experimental builds. These are allowed to fail.
-          #- php: '8.2'
-          #  phpcs_version: 'dev-master'
-          #  wpcs_version: '2.3.*'
-          #  experimental: true
-
```
jrfnl, line 106: Just checking - you are now also removing the "placeholder" for the "future PHP" build - even though it's currently commented out. While this should probably be changed to `8.3` now (and WPCS `dev-develop`) and may potentially still fail, I'm not sure it should be removed.

(The other builds which are being removed are replaced by a build via the matrix, this one isn't being replaced in this PR.)

.github/workflows/test.yml
```diff
@@ -64,46 +65,18 @@ jobs:
       # - phpcs_version: The PHPCS versions to test against.
       #   IMPORTANT: test runs shouldn't fail because of PHPCS being incompatible with a PHP version.
       #   - PHPCS will run without errors on PHP 5.4 - 7.4 on any supported version.
-      #   - PHP 8.0 needs PHPCS 3.5.7+ to run without errors.
+      #   - PHP 8.0 needs PHPCS 3.5.7+ to run without errors, and we require a higher minimum version.
+      #   - PHP 8.1 needs PHPCS 3.6.1+ to run without errors, but works best with 3.7.1+, and we require at least this minimum version.
```
jrfnl, line 69: Do these comments still have value once the minimum supported PHPCS version has been raised to PHPCS 3.7.1 ?

.github/workflows/test.yml
```diff
@@ -9,6 +9,7 @@ on:
     paths-ignore:
       - '**.md'
   pull_request:
+
   # Allow manually triggering the workflow.
```
jrfnl, line 13: Not sure why an extra new line is being added ? `workflow_dispatch` is just another workflow trigger, just like `pull_request`, so this is not a different "section" or anything in the workflow.

(same remark for the other workflow)

jrfnl (commented):
Just left some questions, no biggie.

jrfnl:
As for `2.3.4` or `3.0.0` - semantically, it's allowed in either, so I have no opinion on where it should go (other than my earlier remark that it may be useful to do a last 2.x release before `develop` becomes 3.x).

.github/workflows/test.yml
GaryJones: Yes - I recall that 8.2 wasn't working correctly, though I can't recall why for the moment. It needs some separate testing, and this 8.2 and 8.3 can be added back in as non-comment as needed. Easy enough to find this placeholder again in Git :-) 

https://github.com/Automattic/VIP-Coding-Standards/issues/745 opened to look into this.

.github/workflows/test.yml
GaryJones: To me they do - they explain some of the decisions of why we can test with certain PHP versions and PHPCS versions.

.github/workflows/test.yml
GaryJones: I validated this against a schema for GitHub Workflows, and it didn't like the comment on the next line after `pull_request:` (presumably wanted an indent); adding a clear line space made the validation happy again.

PR #750, opened 2023-03-04
CI: Test Higher PHP versions

### LINKED ISSUE #745: Test against PHP 8.2, and maybe 8.3
PHP 8.2 is released, and we should check how VIPCS works with it, even though WordPress itself doesn't yet support PHP 8.2.

Add `8.2` to the `test.yml` matrix. We'll also want to use `dev-develop` version of WPCS as well.

Also, add `8.3` as an `include`, but allow it to be experimental.

GaryJones:
Fixes #745.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,10 +28,10 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


jrfnl (approved):
LGTM

PR #751, opened 2023-03-05
RestrictedFunctions: Add Warning for filemtime()

### LINKED ISSUE #684: Sniff for use of `filemtime()`

Because of the container-based infrastructure, VIP does not recommend using the function `filemtime()` to populate a value for versioning; each time a container is created, it would result in a different value. It would be great to warn developers of this using the coding standards and have the bot notify in pull requests.

We provide this information in the docs for [file concatenation and minification](https://docs.wpvip.com/technical-references/vip-platform/file-concatenation-and-minification/), but a CS sniff would be a better way to widely share that information.

A warning on the use of `filemtime()` due to the container-based infrastructure.

Other PHP functions return a time based on when an action was taken against a file in the file system.  It would be worth considering if this sniff should apply to these too:

* https://www.php.net/manual/en/function.filectime.php
* https://www.php.net/manual/en/function.filemtime.php
* https://www.php.net/manual/en/function.getlastmod.php
* https://www.php.net/manual/en/function.fileatime.php

```php
function add_js_scripts(){
    wp_register_script(
        'script-name',
        get_template_directory_uri() . '/js/script.js', 
        array(),
        filemtime( get_template_directory() . '/js/script.js' ), 
        false
    ); 
    wp_enqueue_script( 'script-name' );
}
add_action( 'wp_enqueue_scripts', 'add_js_scripts' );
```

```php
function add_js_scripts(){

    $ver = filemtime( get_template_directory() . '/js/script.js' );

    wp_register_script(
        'script-name',
        get_template_directory_uri() . '/js/script.js', 
        array(),
        $ver, 
        false
    ); 
    wp_enqueue_script( 'script-name' );
}
add_action( 'wp_enqueue_scripts', 'add_js_scripts' );
```

None. Even using `filemtime()` outside of the context of enqueuing could be problematic.

GaryJones:
Because of the container-based infrastructure, VIP does not recommend using the function `filemtime()` to populate a value for versioning; each time a container is created, it would result in a different value.

See https://docs.wpvip.com/technical-references/vip-platform/file-concatenation-and-minification/

Closes #684.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
@@ -223,6 +223,13 @@ class RestrictedFunctionsSniff extends AbstractFunctionRestrictionsSniff {
 					'lchown',
 				],
 			],
+			'filemtime' => [
+				'type'      => 'warning',
+				'message'   => 'VIP discourages the use of %s(), as each time a container is created, it would result in a different value for files in that container. See https://docs.wpvip.com/technical-references/vip-platform/file-concatenation-and-minification/ for more info.',
+				'functions' => [
+					'filemtime',
+				],
+			],
 			'stats_get_csv' => [
 				'type'      => 'error',
 				'message'   => 'Using `%s` outside of Jetpack context pollutes the stats_cache entry in the wp_options table. We recommend building a custom function instead.',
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Functions/RestrictedFunctionsSniff.php
```diff
@@ -223,6 +223,13 @@ public function getGroups() {
 					'lchown',
 				],
 			],
+			'filemtime' => [
+				'type'      => 'warning',
+				'message'   => 'VIP does not recommend %s() as each time a container is created, it would result in a different value for files in that container. See https://docs.wpvip.com/technical-references/vip-platform/file-concatenation-and-minification/ for more info.',
```
jrfnl, line 228:
```suggestion
				'message'   => 'VIP discourages the use of %s(), as each time a container is created, it would result in a different value for files in that container. See https://docs.wpvip.com/technical-references/vip-platform/file-concatenation-and-minification/ for more info.',
```

jrfnl (commented):
One phrasing suggestion, other than that, LGTM.

jrfnl:
Open question: in the original issue several other functions are mentioned as potentially problematic well. Should those be included in this PR too ?

jrfnl:
Note: I've changed the milestone to `2.4.0` as this effectively adds a new check to VIPCS, which is something which should be done in a minor, not a patch release.

GaryJones:
https://github.com/Automattic/VIP-Coding-Standards/issues/684#issuecomment-1612967378

PR #755, opened 2023-06-22
CI: Validate the PHPCS installed standards

GaryJones:


jrfnl:
Just checking: you are adding this as a debug step ? As it doesn't actually validate anything (exitcode will always be `0`).

Also FYI (though I'm pretty sure you already know this): the paths to the standards installed are displayed at the end of the `composer install` step, so are already available.

PR #757, opened 2023-06-22
CI: test and quicktest tweaks

GaryJones:
Turn on display_errors for quicktest, and separate out PHPCS and WPCS composer requires into separate steps.


jrfnl:
FYI: for some other standards (and I intend to do the same for WPCS) I have implemented a change in the workflow logic now as there are more dependencies involved. This change leverages the Composer `--prefer-lowest`/`--prefer-stable` options to run the tests against high/low supported versions. We may want to consider implementing that logic here in the future as well.

PR #758, opened 2023-06-24
Some simplifications of sniffs extending the WPCS `AbstractArrayAssignmentRestrictionsSniff`

jrfnl:
### Some simplifications of sniffs extending the WPCS `AbstractArrayAssignmentRestrictionsSniff`

The WPCS `AbstractArrayAssignmentRestrictionsSniff` expects an array from `getGroups()` which includes a `'message'` key.
The `callback()` method subsequently allows to override that message with a higher priority message if needed in specific circumstances.

None of these sniffs, however, have a need to override the message, so the `callback()` method can just use the default behaviour of returning `true` or `false` and use the `'message'` as declared in the group.

### Remove some extraneous comments

... from these same sniffs. These comments are copy/paste documentation from the abstract and not specific to the actual sniff.


GaryJones (approved):
✅

PR #760, opened 2023-08-22
Functions/RestrictedFunctions: remove reference to function which doesn't exist

jrfnl:
PHP does not contain a `delete()` function, so searching for it will only yield false positives.

Ref: https://www.php.net/manual/en/function.delete.php


WordPress-VIP-Go/ruleset.xml
```diff
@@ -7,7 +7,7 @@
 	<rule ref="WordPressVIPMinimum"/>
 
 	<!-- Things that may be incompatible with the VIP Go infrastructure and needs a dev to review -->
-	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_delete">
+	<rule ref="WordPressVIPMinimum.Functions.RestrictedFunctions.file_ops_unlink">
```
GaryJones, line 10: There's already an entry for unlink on line 80, so this entry can be deleted completely.

GaryJones (changes requested):
I'm not worried about any duplicates in test files, but I think the ruleset should avoid duplicate entries.

WordPress-VIP-Go/ruleset.xml
jrfnl: Ow, good catch, thank you! And yes, I agree. While it doesn't do any harm (at this time) having the duplicate in the ruleset, it will make any debugging later more difficult, so let's just remove in both cases (ruleset + tests).

jrfnl: Updated the PR to reflect this.

GaryJones (approved):
Thanks for updating! ✅

PR #761, opened 2023-08-22
Performance/RegexpCompare: remove redundant condition

jrfnl:
This condition will always be `true` if one of the previous conditions is `true`, so is redundant.


GaryJones (approved):
✅

Issue #588, opened 2020-10-12
Bump VariableAnalysis to ^2.9.0 to allow use of the backported options


GaryJones:
## What problem would the enhancement address for VIP?

When template partials are `require`'d from within a class, then the partial may use `$this->` to insert display a value. At PHP runtime, it works, but for PHPCS analysis, it sees a PHP file with an undefined `$this` and so reports it via the VariableAnalysis package. Technically it's correct, and while it could be ignored, it's not practical or desirable to do that.

## Describe the solution you'd like

[VariableAnalysis 2.9.0](https://github.com/sirbrillig/phpcs-variable-analysis/releases/tag/v2.9.0) includes two new options that can control whether these violations are raised. I'd like to have VIPCS require this update in package, so to be able to make use of these options in the default configuration of VIPCS.

VIPCS currently requires `^2.8.3`, so we'd need to bump the minimum version.

## What code should be reported as a violation?

–

## What code should *not* be reported as a violation?

Don't report on `$this`:

```php
<?php
// Example foo.php template.

if ( ! empty( $this->vars['video-id'] ) ) {
	echo '<video...></video>';
}
```

Don't report on `$args` ([new feature in WP 5.5](https://make.wordpress.org/core/2020/07/17/passing-arguments-to-template-files-in-wordpress-5-5/)):

```php
<?php
// Example foo.php template.
 
// Set defaults.
$args = wp_parse_args(
    $args,
    array(
        'class'          => '',
        'arbitrary_data' => array(
            'foo' => 'fooval',
            'bar' => false,
        ),
        ...
    )
);
?>
 
<div class="widget <?php echo esc_html_class( $args['class'] ); ?>">
    <?php echo esc_html( $args['arbitrary_data']['foo'] ); ?>
</div>
```

GaryJones:
Note that the latest VariableAnalysis is 2.11.10, and it includes a lot of the 3.x benefits backported into 2.x.

jrfnl:
Ref: #691

jrfnl:
I've just been looking at this, but this ticket looks like a duplicate of https://github.com/Automattic/vip-go-ci/issues/120 (different repo), which was fixed by a combination of PR #690 (VA configuration) and #691 (VA version bump), both of which were included in the 2.3.3 release.

I have a feeling this ticket should be closed as fixed in 2.3.3. What am I missing ?

GaryJones:
[...]

Issue #600, opened 2020-12-18
Deprecate WordPressVIPMinimum ruleset


rebeccahum:
With all sites migrated to Go, I think we can deprecate https://github.com/Automattic/VIP-Coding-Standards/blob/develop/WordPressVIPMinimum/ruleset.xml and use https://github.com/Automattic/VIP-Coding-Standards/blob/develop/WordPress-VIP-Go/ruleset.xml as the main source of truth.

rebeccahum:
I guess technically, we could deprecate the Go ruleset and move everything on there to the VIPMinimum depending on which way we want to go about things...but I don't think it's necessary to have two rulesets anymore.

GaryJones:
VIPCS contains two standards - `WordPressVIPMinimum`, meant for code that lives on WordPress.com VIP, and `WordPress-VIP-Go`, meant for code that lives on the VIP Go platform. Once customers have migrated their sites to VIP Go (which is now complete), then there's no reason for them to be using `WordPressVIPMinimum` ruleset directly.

The VIP Go standard doesn't have sniffs of it's own - it references sniffs that are under the `WordPressVIPMinimum` namespace and directory, as well as those from WPCS and built-in coding standards.

The relevant part of the VIPCS repo looks like this:
```
.
├── WordPress-VIP-Go/
│   ├── ruleset-test.inc
│   ├── ruleset-test.php
│   └── ruleset.xml
├── WordPressVIPMinimum/
│   ├── Sniffs/
│   │   ├── AbstractVariableRestrictionsSniff.php
│   │   ├── Classes/
│   │   ├── Compatibility/
│   │   ├── Constants/
│   │   ├── Files/
│   │   ├── Functions/
│   │   ├── Hooks/
│   │   ├── .../
│   ├── Tests/
│   │   ├── Classes/
│   │   ├── Compatibility/
│   │   ├── Constants/
│   │   ├── Files/
│   │   ├── Functions/
│   │   ├── .../
│   ├── ruleset-test.inc
│   ├── ruleset-test.php
└── └── ruleset.xml
```

Understanding what severity level and message applies when using the VIP Go standard means checking:

  - The `WordPress-VIP-Go/ruleset.xml`
  - The `WordPressVIPMinimum/ruleset.xml`, 
  - The WordPressVIPMinimum sniff,
  - Potentially a parent (likely WPCS) sniff that the WordPressVIPMinimum sniff extends.

...as each item here overrides the one below it.

[...]

PR #763, opened 2023-08-22
Rulesets: move VariableAnalysis configuration from Go to Minimum

jrfnl:
While looking into #588, I noticed something which seemed odd, so I'm addressing this with this PR.

The configuration which was added in #690 (and addressed #588) was added to the `WordPress-VIP-Go` ruleset, not the `WordPressVIPMinimum` ruleset from which the `WordPress-VIP-Go` ruleset inherits.

I can not think of any pertinent reason why the configuration should be added for one, but not the other and I couldn't find any discussion on this either, so I propose moving the configuration from the `WordPress-VIP-Go` ruleset to the `WordPressVIPMinimum` ruleset.

Ref: https://github.com/sirbrillig/phpcs-variable-analysis/releases/tag/v2.9.0

### DIFF
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -244,14 +244,6 @@
 		<severity>0</severity>
 	</rule>
 
-	<!-- Do not report on undefined variables before require nor in file scope. -->
-	<rule ref="VariableAnalysis.CodeAnalysis.VariableAnalysis">
-		<properties>
-			<property name="allowUnusedVariablesBeforeRequire" value="true"/>
-			<property name="allowUndefinedVariablesInFileScope" value="true"/>
-		</properties>
-	</rule>
-
 	<rule ref="WordPress.DB.SlowDBQuery.slow_db_query_meta_key">
 		<!-- We are silencing this one because VIP Go has a combined index on meta_key, meta_value-->
 		<severity>0</severity>
--- a/WordPressVIPMinimum/ruleset.xml
+++ b/WordPressVIPMinimum/ruleset.xml
@@ -145,6 +145,13 @@
 
 	<!-- Include VariableAnalysis checks -->
 	<rule ref="VariableAnalysis"/>
+	<rule ref="VariableAnalysis.CodeAnalysis.VariableAnalysis">
+		<properties>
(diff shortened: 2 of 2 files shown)


GaryJones:
Nobody should be using `WordPressVIPMinimum` standard at all, so putting it in the `WordPress-VIP-Go` just made it a little more visible to users who looked at the ruleset file (though the full picture does need both ruleset files).

What advantage does having it in the `WordPressVIPMinimum` give?

If the name of the `WordPress-VIP-Go` ruleset didn't change for 4.x, then the plan would be to move everything from `WordPressVIPMinimum` into the newer ruleset, and then eliminate obvious immediate overrides.

jrfnl:
> Nobody should be using `WordPressVIPMinimum` standard at all, so putting it in the `WordPress-VIP-Go` just made it a little more visible to users who looked at the ruleset file (though the full picture does need both ruleset files).
> 
> What advantage does having it in the `WordPressVIPMinimum` give?

Well, some plugin devs may be (= are) using VIPCS outside of the VIP CI context, for those, it maybe useful.

Also, if you look at the rulesets, Minimum basically contains the majority of the "base" config, while Go builds on top of that and customizes.
As this is "base" config, it just feels like something which belongs in Minimum, not necessarily Go (other than if Go would add/change the config).


> If the name of the `WordPress-VIP-Go` ruleset didn't change for 4.x, then the plan would be to move everything from `WordPressVIPMinimum` into the newer ruleset, and then eliminate obvious immediate overrides.

`WordPress-VIP-Go` is not a name for a standard which can contain sniffs (as it contains `-` dashes, which can't be used in a namespace).

If I remember correctly, a potential ruleset reorganisation is being discussedin #600. Probably best to keep the discussion in that issue. I've moved the ticket to the 4.x milestone now for higher visibility/findability.

GaryJones:
> Well, some plugin devs may be (= are) using VIPCS outside of the VIP CI context, for those, it maybe useful.

Even if they are, they should still be using the WordPress-VIP-Go.

In this case, the changes for the variable analysis config is far more targeted to theme developers.

GaryJones (approved):
✅

jrfnl:
> In this case, the changes for the variable analysis config is far more targeted to theme developers.


Good point. Either way, this change doesn't do any harm.

PR #764, opened 2023-08-22
Composer: update VariableAnalysis dependency

jrfnl:
At this time, the minimum supported VA dependency is set at `2.11.1` after PR #691, which was included in VIPCS 2.3.3.

Since then, there have been a bunch of new releases. While the minimum supported version doesn't necessarily have to be updated each time, there is one reason why I believe it would be a good idea to update it:
* v 2.11.10 widened the allowed versions for the Composer PHPCS installer plugin. Without this fix, the Composer plugin cannot update to v 1.0.0 as VA blocks the update.

And if we're updating VA anyway, I suggest we go straight to the latest & greatest as quite a few bugs, most to do with modern PHP support, have been fixed in the mean time.

Refs: https://github.com/sirbrillig/phpcs-variable-analysis/releases


GaryJones (approved):
✅

PR #766, opened 2023-08-22
Various minor doc fixes

jrfnl:


WordPressVIPMinimum/Sniffs/Performance/RemoteRequestTimeoutSniff.php
```diff
@@ -10,7 +10,7 @@
 use WordPressCS\WordPress\AbstractArrayAssignmentRestrictionsSniff;
 
 /**
- * Flag REGEXP and NOT REGEXP in meta compare
+ * Flag use of a timeout of more than 3 for remote requests.
```
GaryJones, line 13:
```suggestion
 * Flag use of a timeout of more than 3 seconds for remote requests.
```

Admittedly not very clear in the code itself.

GaryJones (changes requested):

WordPressVIPMinimum/Sniffs/Performance/RemoteRequestTimeoutSniff.php
jrfnl: Fixed.

GaryJones (approved):
✅

PR #767, opened 2023-08-23
GH Actions: 2 minor simplification

jrfnl:
### GH Actions: minor simplification [1]

... of the bash `date` command in the earlier pulled cache busting.

### GH Actions: minor simplification [2]

Removes the `experimental` key from the matrix as we can just use a plain comparison in the `continue-on-error` key.


GaryJones (approved):
✅

PR #768, opened 2023-08-23
Classes/DeclarationCompatibility: remove redundant code

jrfnl:
The `$functionList` property is only ever written to, never _read_, so all that work walking all the tokens in the complete class scope is redundant (and was done in a highly inefficient way in the first place).

So, removing all code related to the property should improve performance of the sniff a little.


GaryJones (approved):
I wonder if this was meant to be used in a later development.

jrfnl:
> I wonder if this was meant to be used in a later development.

Quite possibly or a left-over of something removed.

I did try to do a trace back, but it came in with the initial import from SVN, so I couldn't trace the history back any further.

PR #774, opened 2023-08-24
Performance/BatcacheWhitelistedParams: remove the sniff

### LINKED ISSUE #613: Remove BatcacheWhitelistedParamsSniff

jrfnl:
* Remove the sniff.
* Remove the related test in the `WordPressVIPMinimum/ruleset-test.inc` file.
* Remove the error silencing from VIP-Go.
* Remove the related test in the `WordPress-VIP-Go/ruleset-test.inc` file.
    This one is a little more involved. Basically the call to `wp_verify_nonce()`, which is being removed, was "silencing" the nonce verification error for other tests as well, most notably for the tests on line 83-85, due to most tests being in the global scope.
    Looking at it more closely, turns out that line 83 wasn't testing what it was supposed to be testing.
    The error which was previously being thrown on line 83 was about the nonce verification being missing, while the test is annotated to be about the `WordPress.Security.ValidatedSanitizedInput[.InputNotSanitized]` error, which wasn't being thrown.
    Adding a nonce verification check on some empty lines above these tests gets rid of the nonce verification errors, but now left line 83 not testing anything at all (as no key is accessed in the superglobal).
    Adding a random key gets us the error which was intended to be thrown on this line, but now also adds the "missing validation" error. IMO, this is correct (better than it was before), so I'm also updating the test expectations for line 83.

Fixes #613


GaryJones (approved):
✅

PR #777, opened 2023-08-24
3.0: start using PHPCSUtils

### LINKED ISSUE #474: PHPCSUtils: Add dependency

Much of the logic in our sniffs does not account for certain code styles, more modern PHP syntax, and misses other edge cases related to whitespace etc.

Rather than having to spend the time working out all of the edge cases of unusually formatted code, writing unit tests, and 

[PHPCSUtils](https://phpcsutils.com/) is a set of utilities to aid developers of sniffs for [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer).

This package offers the following features [taken from the above linked page]:

**Use the latest version of PHP_CodeSniffer native utility functions.**
Normally to use the latest version of PHP_CodeSniffer native utility functions, you would have to raise the minimum requirements of your external PHPCS standard.

Now you won't have to anymore. This package allows you to use the latest version of those utility functions in all PHP_CodeSniffer versions from PHPCS 2.6.0 and up.

**A number of abstract sniff classes which your sniffs can extend.**
These classes take some of the heavy lifting away for a number of frequently occurring sniff types.

**A collection of static properties for often used token groups.**
Collections of related tokens as often used and needed for sniffs.
These are additional "token groups" to compliment the ones available through the PHPCS native PHP_CodeSniffer\Util\Tokens class.

**An ever growing number of utility functions for use with PHP_CodeSniffer.**
Whether you need to split an array into the individual items, are trying to determine which variables are being assigned to in a list() or are figuring out whether a function has a DocBlock, PHPCSUtils got you covered!

Includes improved versions of the PHPCS native utility functions and plenty new utility functions.

These functions are, of course, compatible with PHPCS 2.6.0 up to PHPCS master.

**Test utilities**
An abstract UtilityMethodTestCase class to allow for testing your own utility methods written for PHP_CodeSniffer with ease.
Compatible with both PHPCS 2.x as well as 3.x. Supports PHPUnit 4.x up to 8.x.

**Backward compatibility layer**
A PHPCS23Utils standard which allows sniffs to work in both PHPCS 2.x and 3.x, as well as a number of helper functions for external standards which still want to support both PHP_CodeSniffer 2.x as well as 3.x.

**Fully documented**
To see detailed information about all available abstract sniffs, utility functions and PHPCS helper functions, have a read through the extensive documentation.

## Additional context

This Issue is to summarise what PHPCSUtils is, and to be the point of reference for the PR that adds the dependencies.

Further Issues / PRs will be needed for actually swapping in the [utility functions](https://phpcsutils.com/phpdoc/namespaces/phpcsutils-utils.html).

jrfnl:
### Composer: require PHPCSUtils

By starting to use PHPCSUtils, we make VIPCS less dependent on WordPressCS. It has brings added benefits of more stable, thoroughly tested utility methods, which all support modern PHP.

This adds the dependency and updates relevant documentation.

**Note**: I've _not_ changed the GH Actions workflows to add PHPCSUtils to the matrix (yet). Changes to the workflows will be done in a separate commit.

### VIPMinimum ruleset: add PHPCSUtils requirement

PHPCSUtils does not contain any sniffs, so adding this rule isn't strictly necessary, but by having the rule in the ruleset anyway, if PHPCSUtils is missing, the user will get a descriptive error message during the loading of the ruleset instead of a fatal "class not found" error once the sniffs start running.

Adding this only in the `WordPressVIPMinimum` ruleset is sufficient as the `WordPress-VIP-Go` ruleset includes `WordPressVIPMinimum`.

### AbstractVariableRestrictionsSniff: use PHPCSUtils MessageHelper::addMessage

This is more tested than the WPCS Sniff::addMessage().

### Classes/DeclarationCompatibility: use PHPCSUtils for retrieving name of extended class

Supersedes #734 

Closes #734
Fixes #474

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -64,6 +64,7 @@ When you introduce new `public` sniff properties, or your sniff extends a class 
 ### Pre-requisites
 * VIP Coding Standards
 * WordPress-Coding-Standards
+* PHPCSUtils 1.x
 * PHP_CodeSniffer 3.x
 * PHPUnit 4.x, 5.x, 6.x or 7.x
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -44,6 +44,7 @@ Use `php -v` and `composer show` to get versions.
 | ------------------------ | -------
 | PHP version              | x.y.z
 | PHP_CodeSniffer version  | x.y.z
+| PHPCSUtils version       | x.y.z
 | VIPCS version            | x.y.z
 | WordPressCS version      | x.y.z
 | VariableAnalysis version | x.y.z
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Go to https://docs.wpvip.com/technical-references/code-review/phpcs-report/ to l
 
 * PHP 5.4+
 * [PHPCS 3.7.1+](https://github.com/squizlabs/PHP_CodeSniffer/releases)
+* [PHPCSUtils 1.0.8+](https://github.com/PHPCSStandards/PHPCSUtils)
 * [WPCS 2.3.0+](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/releases)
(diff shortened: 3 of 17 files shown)


GaryJones (approved):
✅

PR #778, opened 2023-08-24
CS: improve use statements

jrfnl:
What with more `use` statements coming in now with PHPCSUtils and WPCS 3.0, it made sense to create some order in these.

### CS: use alphabetically ordered import use statements


GaryJones (approved):
✅

PR #746, opened 2023-01-30
Fix checks for PHP 8.1 and above

GaryJones:
Running `composer test` (or `composer check` after #741 is merged) would lead to several issues under PHP 8.1 and 8.2. This PR fixes those issues.

## Ruleset tests: Add label before test runs

Make it clear which ruleset is being tested, even on a test failure. This just makes it easier to debug.

## Ruleset Tests: Workaround WPCS trim() bug

PHP 8.1 increased the strictness of types used with native functions, and this meant WPCS 2.3.0 had a bug: https://github.com/WordPress/WordPress-Coding-Standards/issues/2035

For the VIPCS ruleset tests with PHP 8.1 or above, it would skip over a bunch of lines in the test file and report other lines as expecting errors or warnings incorrectly.

Running:

```
vendor/squizlabs/php_codesniffer/bin/phpcs --standard=WordPressVIPMinimum --severity=1 WordPressVIPMinimum/ruleset-test.inc -s
```

...shows all the expected violations, but also this for line 1:

> An error occurred during processing; checking has been aborted. The error message was: trim(): Passing null to parameter #1 ($string) of type string is deprecated in .../vipcs/vendor/wp-coding-standards/wpcs/WordPress/Sniff.php on line 1144 (Internal.Exception)

Line 1144 relates to the retrieval of the `minimum_supported_wp_version` config value. Since it appeared to be coming through as `null`, the `RulesetTest.php` now includes setting a trivial value via a command-line argument when PHPCS is called to run the ruleset test.

When WPCS 3.0 is the minimum supported, this change can be reverted since the original bug would have been fixed.

## CS: Work around WPCS trim() bug

PHP 8.1 increased the strictness of types used with native functions, and this meant WPCS 2.3.0 had a bug: https://github.com/WordPress/WordPress-Coding-Standards/issues/2035

The VIPCS coding standards check with PHP 8.1 or above would report incorrect violations from sniffs that included the retrieval of optional command-line arguments. These violations did not appear when running PHPCS with PHP 8.0 or below.

Since the retrieval of the optional command-line arguments appeared to be coming through as null, causing the problems, the command to run PHPCS on the VIPCS codebase now includes the setting of several command-line arguments.

With the `prefixes` enabled, extra sniff behaviour is enabled, and VIPCS does not need to adhere to consistent prefixes (unlike actual WordPress plugin and theme codebases). As such, the PrefixAllGlobals rule is excluded for VIPCS.

[...]

PR #779, opened 2023-08-24
3.0: support WordPressCS 3.0

### LINKED ISSUE #603: Replace WordPress.PHP.StrictComparisons with Universal.Operators.StrictComparisons
In https://github.com/WordPress/WordPress-Coding-Standards/pull/1919, the `WordPress.PHP.StrictComparisons` sniff was replaced with `Universal.Operators.StrictComparisons`. 

We currently reference it here:

https://github.com/Automattic/VIP-Coding-Standards/blob/60ad148168258afccd05ff0429d53a7e810270a5/WordPress-VIP-Go/ruleset.xml#L216-L218

### LINKED ISSUE #727: Make VIPCS PHP 8-friendlier

jrfnl:
### Composer: update WordPressCS + PHPCS dependencies

WordPressCS 3.0.0 has been released and requires a minimum PHPCS version of 3.7.2.

This commit updates the WordPressCS and PHPCS dependencies and the documentation referring to those, in all the relevant places.

Includes:
* Removing the DealerDirect plugin from `require` as it comes with WordPressCS 3.0.0 automatically (via PHPCSUtils) and not having the explicit dependency may prevent conflicts with allowed versions in the future.
* Updating the PHP version on which the CS check for the VIPCS native code is being run. With WPCS 3.0, this check is no longer limited to PHP 7.4.
* Removing a work-around for WPCS < 3.0 vs PHP 8.0
* Updating the VIPCS native `.phpcs.xml.dist` ruleset.

Ref: https://github.com/WordPress/WordPress-Coding-Standards/releases/tag/3.0.0

### WPCS 3.0 | Revert previously applied work-arounds

These are no longer needed with WPCS 3.0.0.

Ref: #746

### Security/EscapingVoidReturnFunctions: switch to using WPCS PrintingFunctionsTrait

Note: this does mean that the sniff will now also support a `public` `customPrintingFunctions` property which can be adjusted in a custom ruleset.

### AbstractArrayAssignmentRestrictions children: adjust for changes in received `$val`

The `$val` parameter received by the `callback()` method will no longer automatically have been stripped of quotes.

This adjusts the `callback()` methods of the sniff which extend the WPCS `AbstractArrayAssignmentRestrictions` sniff to take this into account.

### VIPMinimum ruleset: replace strict comparison sniff

Includes adding PHPCSExtra to the `composer.json` configuration as that is now a direct dependency of VIPCS as VIPCS now uses one of its sniffs in the ruleset.

Fixes #603

### VIPMinimum ruleset: replace assignment-in-condition sniff

... which has been split.

### VIPMinimum ruleset: update excludes for the WP/AlternativeFunctions sniff

... to prevent introducing new duplicate error messages.

### Update ruleset tests: account for new errors from WPCS

* The `WordPress.Security.ValidatedSanitizedInput` sniff will now also examine `$_SESSION` variables.
    As the test is about another sniff, let's just ignore the notices coming from `WordPress.Security.ValidatedSanitizedInput`.

### WordPressVIPMinimum RulesetTest: fix it

To be honest, I'm completely stumped. I can't, for the life of me, think of a reason why the update of the WPCS dependency would suddenly cause all sorts of notices from the `VariableAnalysis` to show, where they previously did not.

* It can't be related to the version update of the `VariableAnalysis` sniff in VIPCS 2.3.4, nor the move of the properties from the Go ruleset to the Minimum ruleset as, in that case, the build for the earlier PR/commit should also have been failing.
* It can't be related to the VIPCS native `VariableAnalysis` sniff being removed as, in that case, the build for the earlier PR/commit should also have been failing.
* VA and WPCS currently have two overlapping CS related dependencies.
    - PHP_CodeSniffer itself, but as they both use the same version, that can not be a reason for the change in the results.
    - The Composer plugin, but that shouldn't influence the scan results, other than if a ruleset would not be registered, but if that were the case (for VA), the tests should have been failing before on line 264, which specifically tests the VA standard triggering a notice (and yes, I checked, it did do so correctly before and after).
* No new parse errors were accidentally introduced into the test files.

As things are, I'm out of ideas of why these warnings are now suddenly showing up, so all I can do is fix them (by selectively silencing the notices from VA).

Note: this is only happening in the Minimum ruleset test as VIPGo silences the `UnusedVariable` notice completely.

Closes #727 (as WPCS 3.0.0 will no longer throw deprecations on PHP 8.x)


GaryJones (approved):
✅

Issue #672, opened 2021-04-21
WPQueryParams.PostNotIn_exclude: False positive outside get_posts context


rebeccahum:
## Bug Description

We started flagging for `exclude` since it's an alias of `post__not_in` in #589. Unfortunately, when used outside of a `get_posts()` or WP_Query context, it's a false positive.

## Minimal Code Snippet

```php
	$users = get_users(
		array(
			'blog_id'        => false,
			'include'        => $include_blog_users,
			'exclude'        => $exclude_blog_users,
		)
	);
```

## Error Code

WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_exclude

## Environment

| Question                | Answer
| ------------------------| -------
| PHP version             | 7.4 
| PHP_CodeSniffer version | 3.6
| VIPCS version           | 2.3

## Tested Against `master` branch?

- [x] I have verified the issue still exists in the `master` branch of VIPCS.
- [x] I have verified the issue still exists in the `develop` branch of VIPCS.

jrfnl:
From https://github.com/Automattic/VIP-Coding-Standards/issues/369#issuecomment-705898613:

> Also note that `exclude` is quite a generic array key name, so I suspect adding this key as one to be detected by this sniff will yield lots of false positives for `exclude` being used in other contexts than `get_posts()`.
> 
> At the same time, hard linking the key detection to `get_posts()` is also not a realistic option as these kind of parameters are often set up outside of the function call / in class properties / class constants etc.

rebeccahum:
I'm wondering if this might be worth changing the messaging to include a manual inspection disclaimer...

PR #780, opened 2023-08-25
Performance/WPQueryParams: defer to the parent sniff

### LINKED ISSUE #594: Move post__not_in sniff to use the grouping in new abstract of WPQueryParamsSniff
We added detection for `exclude` in https://github.com/Automattic/VIP-Coding-Standards/pull/589... but in an attempt to avoid BC-break of error codes, we didn't port over post__not_in to the grouping.

jrfnl:
This commit removes the custom token target + custom logic from this sniff in favour of deferring to the logic in the parent sniff - as discussed in #589.

To that end, the keys which were handled in the custom `process_token()` logic have now been added to the `getGroups()` array.

As the logic for whether or not an error message should be thrown is different for each group, an extra `'name'` key has been added to each group to allow the `callback()` function to distinguish what group the detected key came from.

It also updates the error message being used to better cover both keys being looked for, as well as mention this only applies when the array is passed to `get_posts()` (in an attempt to remove some of the confusion reported in #672).

Note: this is a BC-break as the error codes for the two out of the three existing checks change!
* `WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn` is now `WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_post__not_in`.
* `WordPressVIPMinimum.Performance.WPQueryParams.SuppressFiltersTrue` is now `WordPressVIPMinimum.Performance.WPQueryParams.SuppressFilters_suppress_filters`.
* The `WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_exclude` error code, as introduced in #589, remains the same.

Fixes #594

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Performance/WPQueryParamsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Performance/WPQueryParamsSniff.php
@@ -8,29 +8,16 @@
 
 namespace WordPressVIPMinimum\Sniffs\Performance;
 
-use PHP_CodeSniffer\Util\Tokens;
 use WordPressCS\WordPress\AbstractArrayAssignmentRestrictionsSniff;
 
 /**
  * Flag suspicious WP_Query and get_posts params.
  *
- *  @package VIPCS\WordPressVIPMinimum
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
✅

Issue #713, opened 2022-04-22
NoPagingSniff: False negative when there is no comma after the parameter


rebeccahum:
## Bug Description

There's a false negative when there's no comma after a single item array. For example, this will flag as expected:

```
$args = array(
	'nopaging'       => true, // Bad.
);
```

However, this won't:

```
$args = array(
	'nopaging'       => true // Bad.
);
```

## Minimal Code Snippet

This should flag as an error:
```php
$args = array( 'nopaging' => true );
```

## Error Code

WordPressVIPMinimum.Performance.NoPaging.nopaging_nopaging

## Environment

Use `php -v` and `composer show` to get versions.

| Question                | Answer
| ------------------------| -------
| PHP version             | 7.4.27
| PHP_CodeSniffer version | 3.6.2
| VIPCS version           | 2.3.3

## Tested Against `master` branch?

- [x] I have verified the issue still exists in the `master` branch of VIPCS.
- [x] I have verified the issue still exists in the `develop` branch of VIPCS.

rebeccahum:
@jrfnl Is this an expected side-effect of the `AbstractArrayAssignmentRestrictionsSniff`?

GaryJones:
This seems like an upstream issue. Doesn't seem to be fixed with WPCS `develop` (pre-3.0) branch.

The abstract class is tested with https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Tests/DB/SlowDBQueryUnitTest.inc and https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Tests/WP/PostsPerPageUnitTest.inc and neither of those have an array key and value without a trailing comma. cc @jrfnl

jrfnl:
I agree, this should be reported upstream.

Also note that the `AbstractArrayAssignmentRestrictionsSniff` class is one I've barely ever touched (original was created before my time) as the premise of that sniff is iffy to say the least, so it's an abstract which is likely to have quite a few more flaws.

GaryJones:
Closing in favour of https://github.com/WordPress/WordPress-Coding-Standards/issues/2211.

PR #781, opened 2023-08-25
Performance/NoPaging: add extra tests

jrfnl:
Safeguard the fix which was added to WPCS 3.0.0 in response to WPCS issue WordPress/WordPress-Coding-Standards #2211, which was created to handle VIPCS issue #713.


GaryJones (approved):
✅

PR #782, opened 2023-08-25
GH Actions: minor tweaks to the composer options used

jrfnl:
Only disregard an upper limit to the PHP restrictions.


GaryJones (approved):
✅

Issue #520, opened 2020-07-27
Review the WordPressVIPMinimum.Hooks.AlwaysReturnInFilter sniff


jrfnl:
Review the `WordPressVIPMinimum.Hooks.AlwaysReturnInFilter` sniff for the following in as far as relevant to that sniff:

- [ ] Code style independent sniffing / Correct handling of quirky code
    Typical things to add tests for and verify correct handling of:
    - [ ] Nested function/closure declarations
    - [ ] Nested class declarations
    - [ ] Comments in unexpected places
    - [ ] Variables being assigned to via `list` statements
    - [ ] Multiline text strings
    - [ ] Text strings provided via heredoc/nowdoc
    - [ ] Use of short open tags
    - [ ] Using PHP close tag as end of statement
- [ ] Code simplifications which can be made using PHPCSUtils
- [ ] Sniff stability improvements which can be made using PHPCSUtils
- [ ] Correct handling of modern PHP code
    Typical things to add tests for and verify correct handling of (where applicable):
    - [ ] PHP 5.0 Try/catch/finally (PHP 5.5) and exceptions
    - [ ] PHP 5.3 Namespaced code vs code in the global namespace
    - [ ] PHP 5.3 Use import statements, incl aliasing
    - [ ] PHP 5.3 Short ternaries
    - [ ] PHP 5.3 Closures, incl closure use
    - [ ] PHP 5.4 Short arrays
    - [ ] PHP 5.5 Class name resolution using `::class`
    - [ ] PHP 5.5 List in foreach
    - [ ] PHP 5.5/7.0 Generators using yield and yield from
    - [ ] PHP 5.6 Constant scalar expressions
    - [ ] PHP 5.6 Importing via `use function/const`
    - [ ] PHP 7.0 Null coalesce
    - [ ] PHP 7.0 Anonymous classes
    - [ ] PHP 7.0 Scalar and return type declarations
    - [ ] PHP 7.0 Group use statements
    - [ ] PHP 7.1 Short lists
    - [ ] PHP 7.1 Keyed lists
    - [ ] PHP 7.1 Multi-catch
    - [ ] PHP 7.1 Nullable types
    - [ ] PHP 7.3 List reference assignments
    - [ ] PHP 7.4 arrow functions
    - [ ] PHP 7.4 numeric literals with underscores
    - [ ] PHP 7.4 null coalesce equals
    - [ ] PHP 7.4 Typed properties
[...]

PR #177, opened 2018-04-23
Add filters always return sniff

david-binda:

PR #783, opened 2023-08-25
Hooks/AlwaysReturnInFilter: remove redundant condition

jrfnl:
Given that:
* `$insideIfConditionalReturn` has a default value of `0`;
* And that value is only ever increased;
* The `$insideIfConditionalReturn >= 0` condition will always be `true`.

So this condition can be safely removed, just like the - now unused - assignments to the variable.

The original condition was introduced with the introduction of the sniff in #177. The condition was adjusted in #291, which made the logic redundant.

Looking at the sniff, I believe the intention was to only flag the "return outside condition missing" when not all control structure paths had a `return` statement, but this was never really properly checked as the only control structures taken into account are `if` control structures.

I believe it would be good to improve the sniff to handle more control structures (`switch`, `while` etc) and to not throw the "return outside condition missing" error if all possible paths have a `return` statement, but that is outside the scope of the current PR.

I will add a note to this effect to the review ticket for this sniff - #520.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Hooks/AlwaysReturnInFilterSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Hooks/AlwaysReturnInFilterSniff.php
@@ -223,13 +223,10 @@ class AlwaysReturnInFilterSniff extends Sniff {
 			$functionBodyScopeEnd
 		);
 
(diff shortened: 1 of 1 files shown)


GaryJones (approved):
✅

Issue #729, opened 2022-11-22
`WPQueryParamsSniff` incorrectly flags `get_users()` call


swissspidy:
## Bug Description

When using code like `get_users( [ 'exclude' => 123 ] )`, you'll get a warning like 

```
Using `exclude`, which is subsequently used by `post__not_in`, should be done with caution, see https://docs.wpvip.com/how-tos/improve-performance-by-removing-usage-of-post__not_in/ for more information.
```

This is wrong because this is not a `WP_Query` call and does not use `post__not_in`.

Querying that many users is still not super ideal, but I think this warrants its own sniff that is less confusing.

## Minimal Code Snippet

```php
get_users( [ 'exclude' => 123 ] )
```

## Environment

Use `php -v` and `composer show` to get versions.

| Question                | Answer
| ------------------------| -------
| PHP version             | 8.1.12
| PHP_CodeSniffer version | 3.7.1
| VIPCS version           | 2.3.3

## Tested Against `master` branch?

- [ ] I have verified the issue still exists in the `master` branch of VIPCS.
- [ ] I have verified the issue still exists in the `develop` branch of VIPCS.

swissspidy:
Duplicate of #672

PR #784, opened 2023-08-25
Performance/WPQueryParams: prevent false positives for 'exclude' with get_users()

### LINKED ISSUE #672: WPQueryParams.PostNotIn_exclude: False positive outside get_posts context

jrfnl:
As reported in #672 and #729, the `get_users()` function also takes an array parameter which takes an `'exclude'` key. That key is not our target, so should not be flagged.

This commit adds a hard-coded exception specifically for that situation.

If at a later point in time, more situations which need exceptions would be discovered, this solution can be made more flexible, but for now, there is no need (or insight into where the flexibility should be).

As the `AbstractArrayAssignmentRestrictionsSniff::callback()` method does not have access to the `$stackPtr`, the logic which can be used in the `callback()` is limited. Also see the review notes from upstream WordPress/WordPress-Coding-Standards#2266, which basically already pointed out this exact problem.

To get round this, I'm overloading the `process_token()` method to set a temporary `$in_get_users` property, which can then be read out in the `callback()` method to be used in the actual determination of whether the exception should be made or not.

Includes tests.

Fixes #672

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Performance/WPQueryParamsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Performance/WPQueryParamsSniff.php
@@ -9,6 +9,7 @@
 namespace WordPressVIPMinimum\Sniffs\Performance;
 
 use WordPressCS\WordPress\AbstractArrayAssignmentRestrictionsSniff;
+use WordPressCS\WordPress\Helpers\ContextHelper;
 
 /**
  * Flag suspicious WP_Query and get_posts params.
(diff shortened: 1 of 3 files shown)


GaryJones (approved):
✅

Issue #507, opened 2020-07-22
Review the WordPressVIPMinimum.Classes.DeclarationCompatibility sniff


jrfnl:
Review the `WordPressVIPMinimum.Classes.DeclarationCompatibility` sniff for the following in as far as relevant to that sniff:

- [ ] Code style independent sniffing / Correct handling of quirky code
    Typical things to add tests for and verify correct handling of:
    - [x] Nested function/closure declarations
    - [x] Nested class declarations
    - [ ] Comments in unexpected places
    - [ ] Variables being assigned to via `list` statements
    - [ ] Multiline text strings
    - [ ] Text strings provided via heredoc/nowdoc
    - [ ] Use of short open tags
    - [ ] Using PHP close tag as end of statement
    - [ ] Inline control structures (without braces)
- [ ] Code simplifications which can be made using PHPCSUtils
- [x] Sniff stability improvements which can be made using PHPCSUtils
- [ ] Correct handling of modern PHP code
    Typical things to add tests for and verify correct handling of (where applicable):
    - [ ] PHP 5.0 Try/catch/finally (PHP 5.5) and exceptions
    - [x] PHP 5.3 Namespaced code vs code in the global namespace
    - [ ] PHP 5.3 Use import statements, incl aliasing
    - [ ] PHP 5.3 Short ternaries
    - [ ] PHP 5.3 Closures, incl closure use
    - [ ] PHP 5.4 Short arrays
    - [ ] PHP 5.5 Class name resolution using `::class`
    - [ ] PHP 5.5 List in foreach
    - [ ] PHP 5.5/7.0 Generators using yield and yield from
    - [ ] PHP 5.6 Constant scalar expressions
    - [ ] PHP 5.6 Importing via `use function/const`
    - [ ] PHP 7.0 Null coalesce
    - [x] PHP 7.0 Anonymous classes
    - [ ] PHP 7.0 Scalar and return type declarations
    - [ ] PHP 7.0 Group use statements
    - [ ] PHP 7.1 Short lists
    - [ ] PHP 7.1 Keyed lists
    - [ ] PHP 7.1 Multi-catch
    - [ ] PHP 7.1 Nullable types
    - [ ] PHP 7.3 List reference assignments
    - [ ] PHP 7.4 arrow functions
    - [ ] PHP 7.4 numeric literals with underscores
    - [ ] PHP 7.4 null coalesce equals
    - [ ] PHP 7.4 Typed properties
[...]

PR #799, opened 2023-09-19
Classes/DeclarationCompatibility: sync with WP Core

jrfnl:
While looking at this sniff for something unrelated, I started wondering if the signature definitions were still in line with WP Core.

Turned out they were not, though with the current checks being done in the sniff, this wasn't necessarily problematic (though it should have been, but that's for another PR).

Also see the [additional notes I've added to the review ticket](https://github.com/Automattic/VIP-Coding-Standards/issues/507#issuecomment-1724805522).

Refs:
* https://developer.wordpress.org/reference/classes/wp_widget/wp_widget/
* https://developer.wordpress.org/reference/classes/walker/start_el/
* https://developer.wordpress.org/reference/classes/walker/end_el/
* https://developer.wordpress.org/reference/classes/walker/unset_children/

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Classes/DeclarationCompatibilitySniff.php
+++ b/WordPressVIPMinimum/Sniffs/Classes/DeclarationCompatibilitySniff.php
@@ -40,7 +40,7 @@ class DeclarationCompatibilitySniff extends AbstractScopeSniff {
 				'widget_options'  => [
 					'default' => 'array()',
 				],
-				'constol_options' => [
+				'control_options' => [
 					'default' => 'array()',
 				],
 			],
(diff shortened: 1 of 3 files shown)


GaryJones:
When did those changes happen in WP core, and would this sniff, in theory, need to differentiate for folks running this sniff on a before and after version of WP core?

jrfnl:
> When did those changes happen in WP core, and would this sniff, in theory, need to differentiate for folks running this sniff on a before and after version of WP core?

I know a lot of variable renaming has been happening since PHP 8.0 as, even though WP doesn't support named parameters, it did bring more focus on using better parameter names.

I can do a git blame to figure out exactly when, but to be honest, it's irrelevant as parameter names are not checked by the sniff (as I mention in the description above + in the comments on the other ticket).

So, while this commit improves the information contained in the sniff, it has no impact on the functionality of the sniff.

GaryJones (approved):
✅

PR #800, opened 2023-09-27
Disable VariableAnalysis for phpcbf to speed up by 20%

kkmuffme:
- Disable VariableAnalysis for phpcbf to speed up by 20%
- Also removes default "ruleset", as there's only one rule which is specifically declared, therefore making the ruleset redundant and causing problems with custom rulesets, that only declare the rule (with phpcs-only="true"), but not the ruleset itself, causing this ruleset to load variable analysis unnecessarily

### DIFF
--- a/WordPress-VIP-Go/ruleset.xml
+++ b/WordPress-VIP-Go/ruleset.xml
@@ -218,7 +218,7 @@
 		<type>warning</type>
 		<severity>3</severity>
 	</rule>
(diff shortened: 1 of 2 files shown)


jrfnl:
Why would this sniff get the `phpcs-only="true"` flag, while there are 40+ sniffs in VIPCS itself, which don't, while none of those contain fixers either ?

PR #805, opened 2023-12-02
Switch to PHPCSStandards/PHP_CodeSniffer & raise minimum supported version

jrfnl:
⚠️  **_This is a DRAFT PR on purpose as it references releases which have not yet been tagged. Once the PHPCS 3.8.0 tag is available ~~(which contains the Composer `replace` directive)~~ and the other dependencies have released, this can/should be merged ~~and released ASAP~~._** ⚠️ 

---

### Switch to PHPCSStandards/PHP_CodeSniffer

The Squizlabs repo has been abandoned. The project continues in a fork in the PHPCSStandards organisation.

Ref:
* squizlabs/PHP_CodeSniffer#3932

### Composer: raise the minimum supported PHPCS version to 3.8.0

... for improved PHP 8.2 support.

~~... to prevent end-users from running into trouble with the name change.~~

~~The files in the Composer `vendor/bin` will only be replaced when the `replace...` directive is found and that is only available in the 3.8.0 tag.~~
~~When the files in the Composer `vendor/bin` are not replaced, they will continue to point to the `vendor/squizlabs/php_codesniffer` directory which will no longer exist, leading to fatal "File not found" errors for end-users trying to run PHPCS/PHPCBF.~~

Includes updating references to the PHPCS version whenever relevant throughout the codebase.

### Composer: update various version constraints

... after the tooling has also updated to the PHPCSStandards version of PHPCS.

Refs:
* https://github.com/PHPCSStandards/PHPCSUtils/releases
* https://github.com/PHPCSStandards/PHPCSExtra/releases

### ~~Changelog for the release of version 3.0.1~~

---

Closes #804

PR #809, opened 2024-02-11
Tests: allow for PHPUnit 8/9

jrfnl:
Follow up on #805

PHP_CodeSniffer 3.8.0 now allows for running the tests, which are based on the PHPCS native test suite, with PHPUnit 8 and 9.

This commit updates the package to take advantage of that.

Includes:
* Widening the PHPUnit version requirements.
* Adding the PHPUnit 8+ cache file to `.gitignore`.
* Updating the PHPUnit configuration to make sure that PHP deprecations will always show, even when on a high PHPUnit 9.6 version.
* Updating the info in `CONTRIBUTING`.
* Simplifications to the `quicktest` and `test` workflows. Also, the code coverage "high" run can now be run against PHP 8.3.
* Running PHPStan against PHP `latest` (couldn't previously be done due to the old PHPUnit version).
* Updating the `unit-tests` script to remove the PHP version based toggle, which is no longer needed.

Ref:
* PHPCSStandards/PHP_CodeSniffer#59

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,5 +1,5 @@
 parameters:
-    #phpVersion: 50400 # Needs to be 70100 or higher... sigh...
+    phpVersion: 70100 # Needs to be 70100 or higher... sigh...
     level: 5
     paths:
         - WordPressVIPMinimum
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -95,7 +95,7 @@ When you introduce new `public` sniff properties, or your sniff extends a class 
 * WordPress-Coding-Standards
 * PHPCSUtils 1.x
 * PHP_CodeSniffer 3.x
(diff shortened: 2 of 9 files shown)


phpstan.neon.dist
```diff
@@ -1,5 +1,5 @@
 parameters:
-    #phpVersion: 50400 # Needs to be 70100 or higher... sigh...
+    phpVersion: 70100 # Needs to be 70100 or higher... sigh...
```
GaryJones, line 2: Is the comment still needed here?

GaryJones (approved):

phpstan.neon.dist
jrfnl: As the code base has a minimum PHP version of 5.4, I think it still has value as it explains why the `phpVersion` is not set to `50400`.

PR #826, opened 2024-11-08
Docs: improve consistency and specificity

jrfnl:
## Context

This Monday, [PHPStan 2.0 will be released](https://phpc.social/@OndrejMirtes/113441109253809720).

I've done some preliminary scans with PHPStan 2.0-dev to check if this would have an impact on this codebase.

Based on that, the changes made in commits https://github.com/Automattic/VIP-Coding-Standards/commit/12154a65de39050aae86dcac126e8a6861742999 and https://github.com/Automattic/VIP-Coding-Standards/commit/b95dde72e15c5493a81c93a4ea4577f28ffdf3d7 are necessary to prevent the CI builds from failing once PHPStan 2.0 comes out.

As I was looking through the doc type declarations anyway, I've enhanced a significant number of other `array` types as well, though there is more to be done still.

### Docs: improve consistency and specificity [1]

This fixes up the `@return` type declarations for all `register()` methods.

### Docs: improve consistency and specificity [2]

This fixes up the `@var` type declarations for all `AbstractFunctionParameterSniff::$target_functions` properties.

### Docs: improve consistency and specificity [3]

This fixes up the `@return` type declarations for all `AbstractVariableRestrictionsSniff::getGroups()` methods.

### Docs: improve consistency and specificity [4]

This fixes up the `@return` type declarations for all `AbstractArrayAssignmentRestrictionsSniff::getGroups()` methods.

### Docs: improve specificity

Miscellaneous other doc specificity fixes.

Note: this doesn't fix everything throughout the codebase, but is an iteration to improve things nonetheless.

### DIFF
--- a/WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
@@ -91,7 +91,7 @@ abstract class AbstractVariableRestrictionsSniff extends Sniff {
 	 *  )
 	 * )
 	 *
-	 * @return array
+	 * @return array<string, array<string, string|array<string>>>
 	 */
 	abstract public function getGroups();
 
--- a/WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
@@ -58,7 +58,7 @@ abstract class AbstractVariableRestrictionsSniff extends Sniff {
 	/**
 	 * Returns an array of tokens this test wants to listen for.
 	 *
-	 * @return array
+	 * @return array<int|string>
 	 */
 	public function register() {
 		// Retrieve the groups only once and don't set up a listener if there are no groups.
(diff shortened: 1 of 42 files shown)


WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
```diff
@@ -58,7 +58,7 @@ abstract class AbstractVariableRestrictionsSniff extends Sniff {
 	/**
 	 * Returns an array of tokens this test wants to listen for.
 	 *
-	 * @return array
+	 * @return array<int|string>
```
GaryJones, line 61: In what case would strings be returned in the array here?

GaryJones (approved):

WordPressVIPMinimum/Sniffs/AbstractVariableRestrictionsSniff.php
jrfnl: When a non-PHP native token constant is used.

The value of PHP native token constants are always integers.
The value of the custom PHPCS native token constants are strings, however.

So returning `T_ENUM` may be either an `int` or a `string` depending on the PHP version used to run PHPCS.
As of PHP 8.1 it would be an `int`, but for PHP < 8.1, the token is polyfilled by PHPCS, so it would be a `string`.

Does that help ?

GaryJones: Perfectly!

jrfnl:
FYI: I've just verified (now PHPStan 2.0 has been released) and we're good: https://github.com/Automattic/VIP-Coding-Standards/actions/runs/11775868645

PR #828, opened 2024-11-08
Composer: prevent a lock file from being created

jrfnl:
Composer 1.10.0 introduced a `lock` config option, which, when set to `false` will prevent a `composer.lock` file from being created and will ignore it when one exists.

This is a useful option for libraries such as this where the `lock` file has no meaning.

It also makes life easier for contributors as they don't have to remember that for this repo they should use `composer update` instead of `composer install`. Both will now work the same.

Refs:
https://getcomposer.org/doc/06-config.md#lock

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -51,6 +51,10 @@ jobs:
           ini-values: error_reporting=E_ALL & ~E_DEPRECATED, display_errors=On
           coverage: none
 
+      - name: Enable creation of `composer.lock` file
+        if: ${{ matrix.dependencies == 'lowest' }}
+        run: composer config --unset lock
+
       # Install dependencies and handle caching in one go.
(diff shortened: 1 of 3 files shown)


GaryJones:
Why does the CI/CD unset the lock value (allowing the composer.lock to be created)?

GaryJones (approved):

jrfnl:
> Why does the CI/CD unset the lock value (allowing the composer.lock to be created)?

The reason is that we first use `ramsey/composer-install` to install & cache the dependencies and then only for builds with `matrix.dependencies == 'lowest'` runs a selective `composer update --prefer-lowest ...` for only the CS dependencies.

Without the `lock` file, that second command would run into the following Composer error:
```
Cannot update only a partial set of packages without a lock file present. Run `composer update` to generate a lock file.
``` 

@GaryJones Does that answer your question ?

GaryJones:
It does (and it's what I suspected) - thanks!

PR #830, opened 2025-03-25
GH Actions: use the xmllint-validate action runner + add some extra checks

jrfnl:
### GH Actions: use the xmllint-validate action runner 

Instead of doing all the installation steps for xmllint validation in the workflow, use the ✨ new dedicated `phpcsstandards/xmllint-validate` action runner instead.

Notes:
* This does mean that the `bin/xml-lint` script is now no longer used in CI.
    I've left the script in place as there may be contributors who want to run these checks locally and the script can help them with that, though happy to remove the script if nobody uses it.

Ref: https://github.com/marketplace/actions/xmllint-validate

### GH Actions: add some additional XML validation checks 

... for dev tool files.

### 🆕 Remove the bin/xml-lint script 

... now it is no longer used in CI.


GaryJones:
Fine to remove the script please.

GaryJones (dismissed):

jrfnl:
> Fine to remove the script please.

Done in a separate commit.

GaryJones (approved):

PR #831, opened 2025-03-25
Remove WP CronInterval rule

davidsword:
VIP's cron setup

https://wpvip.com/blog/wordpress-cron/
https://docs.wpvip.com/wordpress-on-vip/cron-control/

can handle frequent heavy cron schedules. On VIP increased job frequency does not equate to slower site performance as the jobs are run async on dedicated batch containers, not front end requests.


jrfnl:
@davidsword Without giving an opinion on this PR, could you please rebase this PR on the current `develop` branch ? Otherwise you will never get a passing build.

GaryJones:
Rather than remove the rule, could we make it 60s instead of 900s?

GaryJones:
Pulled against `develop` instead of `main`, rebased onto the current `develop`, added ruleset test for VIP-Go-Minimum.

jrfnl (commented):
Looking good ✔️ 

The only thing I'm wondering about is VIP-Go vs VIPMinimum.
I've just checked and the VIPMinimum ruleset includes the sniff as well and explicitly sets it to 15 minutes.
https://github.com/Automattic/VIP-Coding-Standards/blob/ac9c5fc9ca04c1d3d9da38f343a9658474673750/WordPressVIPMinimum/ruleset.xml#L46-L51

Is that still correct ?

GaryJones:
Yes - we've not made a start on consolidating these, so while it's not quite aligned with the ultimate end goal, it is aligned with other rules that currently exist in the initial standard.

PR #846, opened 2025-07-18
Security/PHPFilterFunctions: various sniff improvements

### LINKED ISSUE #542: Review the WordPressVIPMinimum.Security.PHPFilterFunctions sniff
Review the `WordPressVIPMinimum.Security.PHPFilterFunctions` sniff for the following in as far as relevant to that sniff:

    - [ ] Multiline text strings
    - [ ] Use of short open tags
    - [ ] PHP 5.3 Short ternaries
    - [ ] PHP 5.4 Short arrays
    - [ ] PHP 5.5 List in foreach
    - [ ] PHP 7.0 Null coalesce
    - [ ] PHP 7.1 Short lists
    - [ ] PHP 7.1 Keyed lists
    - [ ] PHP 7.1 Multi-catch
    - [ ] PHP 7.1 Nullable types
    - [ ] PHP 7.4 arrow functions

Other:
- [ ] Review PHPDoc comments

- [ ] PHP 8.0 Union types
- [ ] PHP 8.0 Named arguments

jrfnl:
### Security/PHPFilterFunctions: improve the tests 

* Test against false positives for calls to methods or namespaced function calls.
* Test against false positives for attribute class using the same name as the function.
* Ensure function import `use` statements are not flagged. We're not interested in those.
* Document that if the filter is passed in dynamically (via a variable or function call), the sniff will stay silent.
* Add some more variations to the pre-existing tests:
    - Non-lowercase function call(s).
    - Fully qualified function calls.
    - Use PHP 7.3+ trailing comma's in a few function calls.

### Security/PHPFilterFunctions: remove redundant strtoupper() calls 

The `isset( $this->restricted_filters[ $parameters[3]['raw'] ] )` check on the array is case-sensitive, so if the filter constant name was not in upper case already, it wouldn't be matched anyhow.

Also note: the case-sensitivity is correct as constants in PHP are case-sensitive (unless explicitly declared as case-insensitive, which these aren't: https://3v4l.org/vq53U#veol).

Includes updating one of the pre-existing tests to document this.

### Security/PHPFilterFunctions: prevent some false positives 

The `'raw'` key in the parameter arrays returned from the `PassedParameters` class contains - as per the name - the _raw_ contents of the parameter.

Since PHPCSUtils 1.0.0-alpha4, the return array also contain a `'clean'` index, which contains the contents of the parameter cleaned of comments.

By switching to using that key, some false positives get prevented.

Includes unit tests demonstrating the issue and safeguarding the fix.

### Security/PHPFilterFunctions: add support for PHP 8.0+ named parameters 

1. Changed the `$target_functions` property to contain information about the target parameter name and position.
2. Adjusted the logic in the sniff to allow for named parameters using the new PHPCSUtils 1.0.0-alpha4 `PassedParameters::getParameterFromStack()` method.
3. The parameter names used are in line with the name as per the PHP 8.0 release.
    PHP itself renamed a lot of parameters in PHP 8.0. As named parameters did not exist before PHP 8.0, the parameter name as per PHP 8.0 (or above) is the only relevant name.
    Also see: php/doc-en#2044
4. Updated the error messages to use the parameter name instead of its position.

As a lot of the logic is now independent of which function is called, this commit also reduces code duplication in the sniff by some logic changes.

Includes additional unit tests.

👉🏻 **_Note: in the context of named parameters, it would be advisable to rename the `MissingSecondParameter` and `MissingThirdParameter` error codes to a dynamic error code using the parameter name instead, but as that would be a BC-break, this will need to wait for the next major (if deemed worth making the change)._**

### Security/PHPFilterFunctions: report restricted filter on the correct token

Throw the warning about the use of a restricted filter on the actual filter constant, not on the function call token. This will more clearly indicate the issue when people use the PHPCS `code` view.

### Security/PHPFilterFunctions: ignore function calls using PHP 5.6+ argument unpacking

This sniff ignores the function call if it cannot reliably be determined if a warning should be thrown.

So, for function calls using PHP 5.6+ argument unpacking, I've elected to also ignore these function calls.

Includes tests covering the change.

### Security/PHPFilterFunctions: document unsupported PHP feature 

The `$options` parameter for `filter_var_array()` and `filter_input_array()` can take either an integer (filter constant) or an array with options, which could include an option setting the filter constant.

At this time, this sniff does not handle an array with options being passed to these functions. Adding support for this will make the sniff much more complicated as PHP supports multiple array formats.

Additionally, if the `$options` parameter is passed as an array, the likelihood that the array is passed in as a variable increases exponentially, so then the next problem would be finding the variable definition and analysing that.

All in all, this is a rabbit hole without end.

Refs:
* https://www.php.net/manual/en/function.filter-var-array.php
* https://www.php.net/manual/en/function.filter-input-array.php

Closes #542

### DIFF
--- a/WordPressVIPMinimum/Sniffs/Security/PHPFilterFunctionsSniff.php
+++ b/WordPressVIPMinimum/Sniffs/Security/PHPFilterFunctionsSniff.php
@@ -9,6 +9,7 @@
 
 namespace WordPressVIPMinimum\Sniffs\Security;
 
+use PHPCSUtils\Utils\PassedParameters;
 use WordPressCS\WordPress\AbstractFunctionParameterSniff;
 
 /**
@@ -26,15 +27,27 @@ class PHPFilterFunctionsSniff extends AbstractFunctionParameterSniff {
 	protected $group_name = 'php_filter_functions';
 
 	/**
-	 * Functions this sniff is looking for.
+	 * Functions this sniff is looking for and information on which parameter to check for in those function calls.
 	 *
-	 * @var array<string, bool> Key is the function name, value irrelevant.
+	 * @var array<string, array{param_position: int, param_name: string}> Key is the function name.
 	 */
 	protected $target_functions = [
-		'filter_var'         => true,
-		'filter_input'       => true,
-		'filter_var_array'   => true,
-		'filter_input_array' => true,
+		'filter_var'         => [
+			'param_position' => 2,
+			'param_name'     => 'filter',
+		],
+		'filter_input'       => [
+			'param_position' => 3,
+			'param_name'     => 'filter',
+		],
+		'filter_var_array'   => [
+			'param_position' => 2,
+			'param_name'     => 'options',
+		],
+		'filter_input_array' => [
+			'param_position' => 2,
+			'param_name'     => 'options',
+		],
 	];
 
 	/**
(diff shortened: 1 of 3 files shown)


WordPressVIPMinimum/Sniffs/Security/PHPFilterFunctionsSniff.php
```diff
@@ -60,30 +78,52 @@ class PHPFilterFunctionsSniff extends AbstractFunctionParameterSniff {
...
-			if ( isset( $parameters[2], $this->restricted_filters[ $parameters[2]['raw'] ] ) ) {
-				$message = 'Please use an appropriate filter to sanitize, as "%s" does no filtering, see http://php.net/manual/en/filter.filters.sanitize.php.';
-				$data    = [ strtoupper( $parameters[2]['raw'] ) ];
-				$this->phpcsFile->addWarning( $message, $stackPtr, 'RestrictedFilter', $data );
+			$message = 'Missing $%s parameter for "%s()".';
+			$data    = [ $param_name, $matched_content ];
+
+			// Error codes should probably be made more descriptive, but that would be a BC-break.
+			$error_code = 'MissingSecondParameter';
+			if ( $matched_content === 'filter_input' ) {
```
GaryJones, line 113: Rather than check against `filter_input`, could this check against `$param_position` instead? In case any future entries are added from the third param position?

jrfnl: As the extension hasn't effectively been [maintained since 2011](https://wiki.php.net/extensions-unmaintained#variable_and_type_related_extensions), I'm a bit sceptical of the idea of additional functions being added to the `Filter` extension.

I'd say, let's leave it as-is. If in the future additional functions would need to be checked, this can always be changed then.

jrfnl: As a side-note: I don't think the error codes should refer to the parameter position at all, but changing that is a BC-break, so would need to wait for a future major release.

See this remark from one of the commit messages:

> 👉🏻 Note: in the context of named parameters, it would be advisable to rename the MissingSecondParameter and MissingThirdParameter error codes to a dynamic error code using the parameter name instead, but as that would be a BC-break, this will need to wait for the next major (if deemed worth making the change).

GaryJones (approved):

WordPressVIPMinimum/Sniffs/Security/PHPFilterFunctionsSniff.php
jrfnl: Oh sod it. As the error code - at this time - is tied to the parameter position, you have a point, so I've changed it.

jrfnl: Darn... merged while I posted this, I'll check how the PR has gone in (with or without the condition change) and if without, will pull the condition change separately.

jrfnl: Without. Opened PR #848 to update this.

PR #895, opened 2026-07-30
Add sniff for usage of DB_NAME or wpdb->dbname

mchanDev:
DB_NAME and $wpdb->dbname are null on the VIP platform


econscript (approved):
LGTM! Tests passed!

GaryJones (approved):

